Issue #35, opened 2026-09-21
Oversized and other body-parser client errors still return 500 (413/415 expected)


mgcronin:
Found while fixing #21 (malformed JSON → 400).

## Defect

Every other body-parser client error still falls through `errorHandler` to the generic 500. Reproduced against `server/src/middleware/error-handler.ts` with `express.json({ limit: "10b" })` and a 20-byte body:

```
500 {"success":false,"error":{"code":"INTERNAL_ERROR","message":"request entity too large"}}
```

body-parser raises these as `http-errors` with `expose: true` and a 4xx `status`, distinguished by `type`:

| `type` | status |
|---|---|
| `entity.too.large` | 413 |
| `encoding.unsupported` | 415 |
| `charset.unsupported` | 415 |
| `request.aborted` | 400 |
| `request.size.invalid` | 400 |

Each is logged as `error-handler: Unexpected error` and returned as 500, which is the same false server-fault signal #21 describes. #21 deliberately mapped only `entity.parse.failed` to stay in scope.

## Acceptance criteria

- [ ] An over-limit JSON body returns 413 with a stable code (the codebase already uses `PAYLOAD_TOO_LARGE` for its own 413s)
- [ ] The other body-parser types above map to their own status with a stable code, never forwarding `err.message`
- [ ] A test through `createApp` sends an over-limit body and asserts 413

Issue #34, opened 2026-09-21
metis-server image is 1,301 MB against the 350 MB budget — api fails on every pull request


mgcronin:
**The `metis-server` image built on a GitHub-hosted runner is 1,301 MB against the 350 MB budget, so the `api` job fails on every pull request** at *Verify image size budget* — after every test has passed.

```
IMAGE                             SIZE (MB)
metis-server:ci-35667839887            1301
metis-ui:ci-35667839887               206.7
metis-embeddings:ci-35667839887        588.8   (sidecar — exempt)
FAIL: metis-server:ci-35667839887 = 1301 MB exceeds 350 MB
```

(First seen on #33, whose change touches no Docker inputs.)

## Background

`docs/OPERATIONS.md` → *Container Image Sizes* documents the amd64 server image at **~329 MB** under a **350 MB** gate. The old CI ran on an arm64 self-hosted runner, where native arm64 binaries inflated the image, so the gate there was set to 1,200 MB as a regression guard; `build-images.yml` was meant to be the authoritative amd64 check. The move to hosted runners (#1) restored the documented 350 MB gate on amd64 — and this is the first run where an amd64 build reached the gate at all (earlier runs ran out of disk first, see #3).

So either the image has regressed ~4× since the documented figure, or the figure was never measured on amd64. Both mean the documentation and the gate disagree with reality.

## To do

- [ ] Break down the 1,301 MB by layer (`docker history`, `dive`) and identify the largest contributors — likely candidates are dev dependencies in the runtime stage, native platform binaries pulled for all platforms, and the Copilot CLI
- [ ] Reduce the image toward the documented budget, **or** record a measured amd64 budget with its rationale — do not simply raise the number to make CI pass
- [ ] Update *Container Image Sizes* in `docs/OPERATIONS.md` with measured amd64 figures
- [ ] `api` passes on a pull request with no Docker-relevant changes

PR #36, opened 2026-09-21
fix: Resolve #14, #21 — double-encoded request bodies, malformed JSON 500s, literal %d in logs

### LINKED ISSUE #14: Requirement edits, comments, replies and assignments all return 500 — request bodies are JSON-encoded twice
**Approving a requirement from the analysis page returns HTTP 500 every time.** The same defect breaks every write in the requirement collaboration API.

## Cause

`apiFetch` (`ui/src/lib/api-client.ts`) serialises `options.body` itself (`JSON.stringify(body)`). Six call sites in `ui/src/lib/collaboration-api.ts` pass `body: JSON.stringify(...)`, so the request body becomes a JSON *string containing JSON*:

```
"{\"reviewStatus\":...}"
```

The server's strict JSON parser rejects it (`Unexpected token '"'`).

Affected (`collaboration-api.ts`): requirement update (`PUT /requirements/:id`), create comment thread, reply, edit comment, assignments — lines around 73, 92, 100, 108, 131, 153.

Two further call sites pre-stringify a body and should be checked for the same pattern: `ui/src/lib/discussions-api.ts` (~156) and `ui/src/lib/analysis-api.ts` (~576).

## Acceptance criteria

- [ ] No caller passes a pre-serialised string to `apiFetch`
- [ ] A test asserts the request body `apiFetch` sends for `requirementUpdateApi.update` is a JSON object, not a JSON string
- [ ] Consider making `apiFetch` reject (in development) a `body` that is already a string, so this cannot recur silently
- [ ] Approving a requirement from the analysis page succeeds

### LINKED ISSUE #21: Malformed JSON request bodies return 500 instead of 400; logger prints literal %d placeholders
Two small defects seen in the server log during a walkthrough.

## 1. Malformed JSON → 500

A request with an invalid JSON body (body-parser `SyntaxError`) is logged as `error-handler: Unexpected error` and returned as **500**. It is a client error and should be **400** with a clear message; 500s also page on-call and hide real server faults.

- [ ] body-parser syntax errors map to 400 with a stable error code
- [ ] A route test sends a malformed body and asserts 400

## 2. printf-style placeholders are not interpolated

Log lines are written with the placeholders literally, e.g.:

```
(requirements-extractor): Extracting structured requirements from input (%d chars)
(requirements-extractor): Extracted %d requirements (%d ambiguities, %d evidence needs)
```

- [ ] Replace printf-style messages with interpolated strings or structured fields, so the numbers are logged

mgcronin:
## Description

Closes #14
Closes #21

### #14: request bodies were JSON-encoded twice
`apiFetch` (`ui/src/lib/api-client.ts`) JSON-encodes `body` itself. Six call sites in `ui/src/lib/collaboration-api.ts` passed `JSON.stringify(...)`, so requirement update/approve, create comment (requirement and artifact), reply, edit comment and assign all sent a JSON *string* and got a 500. Those six sites now pass plain objects.

- **`apiFetch` now throws a `TypeError` outside production when `body` is already a string.** A caller that makes the same mistake fails in its first unit test or dev run instead of reaching the server. Production behaviour is unchanged.
- **The two other files the issue named are correct, so they are unchanged.** `discussions-api.ts:156` and `analysis-api.ts:576` go through `streamFetch`, which passes `RequestInit` straight to `fetch`, so the `JSON.stringify` there is needed. The same holds for `ai-client.ts`, `plugins-api.ts`, `test-coverage-api.ts` and `admin/auth/page.tsx`: each calls `streamFetch` or raw `fetch`.
- **Why the existing tests missed it:** every collaboration test mocks `apiFetch` itself, so none of them could see the bytes actually sent. The new test runs the real `apiFetch`, stubs only `fetch`, and decodes the body once, the way the server's strict `express.json()` does.

### #21 (1): malformed JSON now returns 400, not 500
`errorHandler` maps body-parser's `SyntaxError` (`type === "entity.parse.failed"`) to **400 `INVALID_JSON`** with a fixed message.
- The parser's own message quotes part of the request body, so that message is not sent to the client and is not logged. For example, a malformed login body would otherwise put part of a password into the log.
- A `SyntaxError` raised by server code (for example `JSON.parse` of stored data) still returns 500.

### #21 (2): literal `%d` / `%s` in logs
The winston logger has no `format.splat()`, so all **44** printf-style calls in `server/src` (17 files, not only the two named in the issue) printed the placeholder and dropped the value.
- Every value is now passed as structured meta.
- I chose this over adding `splat()`: `splat()` interpolates values into `message`, and `redactInfo` deliberately skips `message`, so values would have bypassed redaction.
- `web-research-augmenter`'s "Invalid proxy URL" line still does not log the URL, because it can carry `user:password@` credentials.
- A source-scan test now fails the build if a new printf-style log call is added.

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

## Testing

**Gate (CLAUDE.md):**
- `pnpm lint`, `pnpm typecheck` and `pnpm changelog:verify` all pass.
- `pnpm test` passed in full on the first run.
- A second full run after a small follow-up edit hit a 5 s timeout in `scripts/lib/gate-cwd-anchoring.test.mjs`. That file is not touched here, and it passes on its own (4.85 s against its 5 s limit). A concurrent agent's run on the same machine hit the same timeout. I then re-ran server and ui: 1101 and 351 files passed, 0 failed.

**Tests that fail before each fix** (checked by reverting the source and keeping the tests):

| Mutation | Went red |
|---|---|
| `ui/src/lib/{api-client,collaboration-api}.ts` reverted to `origin/main` | all 6 collaboration request-body tests + the string-body guard test |
| collaboration call sites reverted, guard kept | all 6 collaboration request-body tests |
| `assertBodyNotPreSerialised` call removed | "rejects an already-serialised string body outside production" |
| guard ignores `NODE_ENV` (throws in production too) | "does not throw on a string body in production" |
| `server/src` reverted to `origin/main` for the log fix | requirements-extractor record test, claim-extractor record test, drift guard (44 offenders) |
| `error-handler.ts` reverted | 2 unit tests (truncated body, double-encoded body) + 2 `createApp` route tests |
| client message = `err.message` | "does not echo any fragment…", "maps a truncated JSON body…" |
| `isJsonParseError` = `err instanceof SyntaxError` (no `type` check) | "leaves a server-side SyntaxError as a 500" |
| `error: err.message` added to the warn log | "does not write any fragment of the rejected body to the server log" |

The log tests go through the **real** logger: they attach a winston `Stream` transport and parse the JSON it emits, rather than mocking `createChildLogger`.

**Coverage.** Per-file line coverage is the same before and after, except `error-handler.ts`, where branch coverage went from 86.84 to 88.09.

| Scope | Before | After |
|---|---|---|
| `ui/src/lib/api-client.ts`, `collaboration-api.ts` | 100 / 100 lines | 100 / 100 lines |
| 18 touched `server/src` files, lines (all) | 93.63% | 93.66% |
| Statements on changed lines (UI) | — | 12/12 (100%) |
| Statements on changed lines (server) | — | 74/90 (82.2%) |

The uncovered changed statements are mostly the timing-dependent stream paths in `copilot-provider.ts`, which had no coverage before this change either. That file is at 70% overall, which is pre-existing.

**Security (OWASP):**
- No dependencies added or changed.
- The client error body is a fixed string. The request body is kept out of both the response and the log, and there is a test for each.
- The log conversion moves values into meta, where the redactor sees them.
- One change to note for review: some error-message strings that the old calls silently dropped (for example a search provider's `fetch` error) are now logged under `error`.

**Live walkthrough:** not done. The dispatch forbids touching the dev server on :3000/:4000 and running `scripts/restart.sh`. Instead I checked the approve path against the code:
- The UI's `review.mutate` sends `{ reviewStatus, version }` to `PUT /api/requirements/:id`.
- The handler writes `reviewStatus` through `updateRequirementWithHistory`, and it is in `TRACKED_FIELDS`.
- The analysis detail read (`analysis-service.ts` ~1238) reads the same `reviewStatus` column.

So there is no write/read mismatch behind the encoding fix.

## Found along the way
- #35: an oversized body (`entity.too.large`) and the other body-parser client errors still return 500. I reproduced it with a 10-byte limit, and it returned `500 INTERNAL_ERROR "request entity too large"`. I filed it separately rather than widening #21.

## Changelog
`.changes/unreleased/14-collaboration-double-encoded-bodies.md`, `.changes/unreleased/21-malformed-json-400-and-log-values.md`. No version bump.


mgcronin:
CI status for this PR: every check passes except `api` / `api-outcome`, and that failure is not caused by this change.

- **What failed:** only the `api` job's **Verify image size budget** step (`metis-server = 1301 MB exceeds 350 MB`). This is a real failure, not a concurrency supersede.
- **The code gates inside `api` passed:** Lint, Typecheck and Test all succeeded, as did the container image build that runs before the size check (run 35669410304).
- **It already fails on other branches:** the same 1301 MB failure appears on `fix/pinned-lookup-all-addresses` (run 35667839887), and `api` fails on every Dependabot PR. It is tracked as #34.
- **This PR cannot have caused it:** it changes no dependencies, lockfile or Dockerfile.

I have not re-run it. A re-run cannot fix a size budget.

mgcronin (commented):
Recommendation: merge with follow-ups (5 non-blocking findings). The code has no blocking findings. Merging still depends on the `api` / `api-outcome` red check (#34, explained below).

This is an automated review, not a human sign-off. The reviewing `gh` identity is also the PR author, so GitHub cannot record an approval or change request from it. A human owns the merge decision.

## Code Review Summary: PR #36 (#14, #21)

### Requirements: PASS (one criterion checked by tests only)
- [x] #14: none of the six `collaboration-api.ts` call sites passes a pre-serialised string any more. I checked every other `JSON.stringify` body in `ui/src`: `discussions-api.ts:156`, `analysis-api.ts:576`, `ai-client.ts:229` and `plugins-api.ts:64` go through `streamFetch`; `admin/auth/page.tsx` and `test-coverage-api.ts:292` use raw `fetch`. None of them is double-encoded.
- [x] #14: a test asserts that the body `requirementUpdateApi.update` sends decodes to an object. It runs the real `apiFetch` with only `fetch` stubbed.
- [x] #14: the dev-mode guard is in `apiFetch`.
- [~] #14: "Approving a requirement from the analysis page succeeds." Not checked in the running app (the dispatch forbade touching :3000/:4000). The wire body is tested, and `analysis/page.tsx:258` calls the fixed `requirementUpdateApi.update`.
- [x] #21: the body-parser syntax error returns 400 `INVALID_JSON`; tested with the unit handler and through `createApp`.
- [x] #21: printf placeholders fixed at all 44 sites. My own `grep -rnE "%[sdifjoO]"` over `server/src` and `packages` finds none left in a log call.

### Security: FINDINGS (Low)
| Severity | Count | Details |
|---|---|---|
| Low | 1 | `web-research-augmenter.ts:311`: a newly logged `err.message` can contain the Google CSE API key |

Good choices: the parser's message (which quotes part of the body) is kept out of both the response and the log; the proxy URL is deliberately not logged; values go into structured meta, which the redactor covers, instead of `splat()`, which it would skip. None of the new meta keys matches `SENSITIVE_KEY_PATTERNS` in a way that would hide it.

### Code quality: GOOD
The changes are small and follow the existing patterns. `isJsonParseError` keys on `type === "entity.parse.failed"` rather than on any `SyntaxError`, which is correct.

### Performance: CLEAN

### Tests: SUFFICIENT. I checked that each can fail by reverting only the production code:
- `error-handler.ts` reverted: 4 unit tests and 2 `createApp` route tests went red.
- Client message set to `err.message` and the check loosened to any `SyntaxError`: the no-echo, truncated-body and server-SyntaxError-stays-500 tests went red.
- `collaboration-api.ts` reverted with the guard kept: 6 of 6 wire-body tests went red.
- `collaboration-api.ts` and `api-client.ts` both reverted: those 6 plus the guard test went red.
- `requirements-extractor.ts` and `claim-extractor.ts` reverted: both behaviour tests went red, and the drift guard named all 3 offending lines.

The author's mutation claims hold. I did not re-measure coverage; the author reports 82.2% of changed server statements covered, against the 80% floor.

### CI
Every check passes except `api` / `api-outcome`. Inside `api`, Lint, Typecheck, Test and the image build all pass. Only `Verify image size budget` fails, with `metis-server = 1301 MB exceeds 350 MB`. I confirmed the same 1301 MB on unrelated run 35667839887 (`fix/pinned-lookup-all-addresses`). This PR touches no dependency, lockfile or Dockerfile. It is #34 and not caused by this change, but it is a genuine failure (not a supersede), and CLAUDE.md says never to merge with genuinely failing checks. Clearing it or explicitly overriding it is the caller's decision.

### Documentation: UP TO DATE
There are two fragments under `.changes/unreleased/`, one per issue; `changelog` CI passes and no version is bumped. The error-code catalogue has no separate doc that `INVALID_JSON` would need to be added to.

### Scope of this review
I read the full diff and the surrounding code for `api-client.ts`, `error-handler.ts`, `app.ts` (parser order, and the `reconciliationJson` pass-through) and `logger.ts` (redaction). In a separate worktree at e3cd097 I ran the project gate: `pnpm lint` passed, `pnpm typecheck` passed, and `pnpm test` exited 0 (server 1101 files passed, ui 351, scripts 54, shared 36; 0 failed). I then ran the mutations listed above. I did not use the running app.

server/src/lib/analysis/web-research-augmenter.ts
```diff
@@ -297,7 +308,7 @@ export class GoogleCseProvider implements WebSearchProvider {
         snippet: r.snippet ?? "",
       }));
     } catch (err) {
-      log.warn("Google CSE search error: %s", (err as Error).message);
+      log.warn("Google CSE search error", { error: (err as Error).message });
```
mgcronin, line 311: **Low security, low confidence of being reachable.** Before this PR, this `err.message` was silently dropped because the logger has no `splat()`. Now it is logged, and the redactor does not look inside values. Node's `fetch` puts the URL in its parse-error message: I ran `fetch("http://x y/?key=SECRETKEY")` on Node 22.22 and got `"Failed to parse URL from http://x y/?key=SECRETKEY"`. The CSE URL carries `key=<apiKey>` in the query string. It is hard to reach in practice, because `baseUrl` defaults to a constant and the key is `encodeURIComponent`'d. But this is the one site where the new `error` field can carry a credential. Suggest logging `err.name` or `(err as Error & {cause?: unknown}).cause`, or removing `key=` from the message, as you already did for the proxy URL at line 125. The other newly logged `error` fields (LLM/provider failures in ambiguity-grounding and clarification-dialog) look fine.

ui/src/lib/api-client.ts
```diff
@@ -225,6 +225,23 @@ async function doFetch<T>(path: string, options: ApiFetchOptions): Promise<DoFet
...
+/**
+ * #14 — `apiFetch` JSON-encodes `body` itself, so a caller that passes
+ * `JSON.stringify(x)` sends a JSON *string containing JSON*
+ * (`"{\"a\":1}"`), which the server's strict parser rejects. Every
+ * requirement-collaboration write shipped that way and returned 500. Fail
+ * loudly outside production so the mistake surfaces in the first dev run or
+ * unit test instead of as a server error; production behaviour is unchanged.
+ */
+function assertBodyNotPreSerialised(path: string, body: unknown): void {
+  if (typeof body === "string" && process.env.NODE_ENV !== "production") {
```
mgcronin, line 237: nit, Low: the guard catches the #14 shape (a string) but not the other bodies `apiFetch` would silently corrupt. `FormData`, `Blob`, `URLSearchParams` and `ArrayBuffer` all go through `JSON.stringify` and come out as `"{}"`. I found no current caller that passes one, so this is optional hardening, not a defect. Something like `body instanceof FormData || body instanceof Blob || body instanceof URLSearchParams` would close the same class of silent-corruption bug.

server/tests/logger.printf-placeholders.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  return out;
+}
+
+/**
+ * Find `<ident>.<level>(` followed (across newlines) by a string literal whose
+ * text contains a printf token. Returns
+ * `file:line  message` strings so a failure names every offender.
+ */
+function findPrintfLogCalls(source: string, file = "<source>"): string[] {
+  const call =
```
mgcronin, line 145: Low: the drift guard is a good addition, and its self-test proves the scanner can match. Two gaps are worth knowing about. (1) It only matches `<ident>.<level>("literal…"`, so `logger.log("info", "… %d", n)` gets past it. (2) The underlying defect is "positional args are dropped", not "placeholders are printed": `log.warn("failed:", err.message)` loses `err.message` just as surely and contains no `%`. I grepped `server/src` for that second shape and found nothing today, so this is not a current defect. Consider flagging any log call whose second argument is not an object literal or identifier-of-object. Optional.

ui/src/lib/collaboration-api.ts
```diff
@@ -150,7 +150,7 @@ export const requirementUpdateApi = {
       `/requirements/${requirementId}`,
       {
         method: "PUT",
-        body: JSON.stringify(patch),
+        body: patch,
```
mgcronin, line 153: Low, requirement verification: #14's last criterion, "Approving a requirement from the analysis page succeeds", is covered by the wire-body test and by reading the code (`analysis/page.tsx:258` calls this function with `{reviewStatus, version}`), not by an end-to-end run. That is reasonable given the dispatch forbade touching the dev server. Please do a click-through or a Playwright spec after merge so the 500 is confirmed gone in the real app.

server/src/middleware/error-handler.ts
```diff
@@ -127,6 +139,28 @@ export const errorHandler: ErrorRequestHandler = (err, req, res, _next) => {
...
   }
 
+  // #21 — a request body the JSON parser rejects is the CLIENT's error. The
+  // body-parser `SyntaxError` carries `type: "entity.parse.failed"` and
+  // `status: 400`; without this branch it fell through to the 500 below, was
+  // logged as "Unexpected error" and hid genuine server faults. The parser's
+  // own message quotes a fragment of the request body, so it is neither
+  // echoed to the client NOR logged: a malformed login body
+  // would otherwise put a fragment of a password into the log.
+  if (isJsonParseError(err)) {
```
mgcronin, line 149: Informational: correct and well-scoped. Keying on `type === "entity.parse.failed"` rather than any `SyntaxError` keeps server-side `JSON.parse` failures at 500, and the mutation test proves it. The route-scoped `reconciliationJson` parser forwards its non-413 errors through `next(error)`, so it gets the same 400. The other body-parser client errors (`entity.too.large`, `encoding.unsupported`, `charset.unsupported`, `entity.verify.failed`) still return 500. That is #35, correctly left out of this PR's scope.

PR #37, opened 2026-09-22
fix: Resolve #15, #19 — run every tool call in a reply; report analyses that verified nothing

### LINKED ISSUE #15: Agent loop cannot parse several tool calls in one model reply — chat shows raw tool markup and the code agent verifies nothing
**When a model requests more than one tool in a single reply, METIS executes none of them.** In chat the raw tool markup is shown to the user as the answer; in analysis the code agent's investigation ends after one tool.

## Observed

Project chat, question about DNS-rebinding protection. The model replied:

```
<tool_calls>
  <tool_calls>{"tool": "search_code_symbols", "args": {"query": "DNS rebinding SSRF ...", "limit": 15}}</tool_calls>
  <tool_calls>{"tool": "search_code_symbols", "args": {...}}</tool_calls>
  <tool_calls>{"tool": "search_code_graph", "args": {"query": "safeFetch", ...}}</tool_calls>
</tool_calls>
```

That text was rendered as the assistant's answer.

In a four-agent requirements analysis, the code agent ran one `list_files` call, its next reply was not recognised as a tool call, the loop forced a final-answer retry, and all 16 requirements came back `could-not-verify`.

## Cause

`parseToolCall` in `server/src/lib/analysis/agent-loop.ts` handles one JSON object — direct, fenced, or "first `{` to last `}`". With several objects, first-`{`-to-last-`}` spans all of them and is not valid JSON, so the reply is classified as a final answer. `<tool_calls>` wrappers are not recognised at all.

Asking for several tools at once is normal behaviour for current non-Claude models, so this blocks every such provider from the agentic paths.

## Acceptance criteria

- [ ] Consecutive JSON objects, a JSON array of calls, and a `<tool_calls>` wrapper are each parsed into an ordered list of calls
- [ ] Each call is executed in order and every result is returned to the model on the next turn
- [ ] A reply that looks like a tool call but cannot be parsed is never rendered to the user as an answer
- [ ] Parsing stays linear-time on untrusted model output (see the ReDoS notes already in `parseToolCall`)
- [ ] Tests for each encoding, plus a chat-route test that no tool markup reaches the rendered answer

### LINKED ISSUE #19: Analysis health reports "not degraded" when the code agent verified nothing
**An analysis whose code agent made one tool call and verified 0 of 16 requirements is recorded as healthy.** The run's metadata:

```json
"capability": { "codeAgentFailed": false, "codeAgentDegraded": false, "codeRetrievalDegraded": false },
"retrieval":  { "successfulSearches": 1, "totalCalls": 1, "requirementCount": 16,
                "starved": false, "degraded": false }
```

Meanwhile the code agent's own summary said: "Investigation was cut short after a single `list_files` call… none of the 16 requirements could be grounded in code", and every finding was `could-not-verify` at confidence 0.1.

## Why it matters

The health signal exists to warn the user that code grounding failed. Here it reported the opposite, so the analysis page gave no indication that none of its requirements were checked against the code. (Root cause of the agent's failure: the multi-tool-call parsing issue filed separately.)

## Acceptance criteria

- [ ] A run where most requirements are `could-not-verify`, or where retrieval calls per requirement fall below a threshold, is marked degraded/starved
- [ ] The analysis page surfaces that state prominently
- [ ] A test with one search for many requirements asserts `starved: true`

mgcronin:
Closes #15
Closes #19

Both issues came from the same walkthrough. A model asked for several tools in one reply, the loop ran none of them (#15), and the health record still said the run was fine (#19).

## #15: several tool calls in one reply

**Cause:** `parseToolCall` accepts exactly one JSON object. When a reply holds two objects, its "first `{` to last `}`" slice covers both, fails to parse, and the reply is treated as a final answer. `<tool_calls>` wrappers were never recognised.

**Fix** (`server/src/lib/analysis/agent-loop.ts`):
- **`parseToolCalls(response, knownTools)`** returns the calls in order. It accepts consecutive objects, a JSON array (fenced or not), and `<tool_calls>` wrappers, flat or nested the way DeepSeek emits them. It makes one linear bracket scan (string- and escape-aware, active only inside a value) and runs one `JSON.parse` per top-level span. There is no regex. A `findings` answer anywhere in the reply still makes the whole reply the final answer. **`parseToolCall` itself is unchanged**, so the #1244 differential and fuzz tests still pin it. A single-call reply returns `[parseToolCall(...)]`, the same result as before.
- **`runAgentLoop`** runs every call in order and sends all the results back in one user message on the next turn. It checks for abort between calls. A single call produces the same message as before, byte for byte, so prompt caching is unaffected.
- **Per-reply cap:** `MAX_TOOL_CALLS_PER_REPLY = 8`. The reply is untrusted, and the turn cap limits how many replies a run gets but not how many calls one reply can ask for. Calls past the cap are skipped, and the model is told which ones.
- **`looksLikeToolCallMarkup` / `isToolCallReply`** catch tool markup that does not parse: a truncated `{"tool"…`, a fenced or array prefix, or a `<tool_call(s)>` tag followed by JSON or another tag. Such a reply is sent to the final-answer retry and then to the brace-free fallback. It never reaches the user. A tag that is only *mentioned* in prose (for example, an answer explaining the `<tool_calls>` format) is not flagged.
- `classifyFinalAnswer` (#1217 triage) and `isJsonFinalAnswer` (`agentic-degradation.ts`) now recognise multi-call replies. Before, `isJsonFinalAnswer` accepted a multi-call reply as a JSON answer, because `extractJsonObject` pulled out its first object.

## #19: health reported "not degraded" when nothing was verified

**Cause:** the #773 run-level threshold only asks whether retrieval worked at least once, and one working search satisfies it. That is the right question for the **verdict** gate, which has to stay scale-free: a pass-wide quota once made a confirmed gap mathematically impossible at N≈30. But the **report** used the same record.

**Fix** (`retrieval-health.ts` + `orchestrator.ts`):
- **`assessInvestigationCoverage`** runs *after* a pass's verdicts are gated. It only changes the persisted, display-side record. `passHealth`, which the verdicts read, is untouched.
  - **`starved`** is set when the pass made fewer than `MIN_SEARCHES_PER_REQUIREMENT = 0.25` code-retrieval calls per requirement. The incident had 1 call for 16 requirements.
  - **`degraded`** is set when the pass is starved, or when more than `MAX_UNVERIFIED_REQUIREMENT_SHARE = 0.5` of its requirements came back only `could-not-verify`. "Most" means more than half, so exactly half is not flagged.
  - The new optional field `unverifiedRequirements` is recorded when it is above 0 and summed by `mergeRetrievalHealth`.
- A `degraded` record raises `codeRetrievalDegraded`, which is the existing `code-retrieval-degraded` warning at the top of the analysis page. The gap report's retrieval panel now also shows "far fewer code searches than requirements" and "N of M requirements could not be verified".

**Decisions to review:**
1. **Budget-exhausted passes are exempt** from both new rules; `exhausted` is still reported. #1236 deliberately separated exhaustion from starvation, and an existing test (`could-not-verify-pipeline.test.ts`, exhausted suite) asserts `starved: false, degraded: false` for a cut-short pass. The incident was not exhausted: the loop ended because the multi-call reply was misread as an answer.
2. **Unverified is counted conservatively.** A requirement counts only when *every* finding bound to it is `could-not-verify`. Requirements with no findings, and findings with a null verdict, are not counted, so healthy runs stay quiet.
3. **Side effect:** a pass that makes **zero** code searches and relies on the #729 seed is now reported `starved`/`degraded`. Its verdicts are unchanged. That is what #19 asks for, but it will show the warning on seed-only runs that used to be silent.
4. The 0.25 floor is set just under the existing N=20 scale test (5 searches for 20 requirements passes; 4 does not) and far below funded runs, which get 2 turns per requirement.

## Tests: mutation-verified

New: `multi-tool-calls.test.ts` (53 tests: encodings, safety, loop execution and order, cap, abort, markup never returned, 7 ReDoS shapes × {200 KB under 100 ms, linear scaling}), `ai-routes-multi-tool-call.test.ts` (real `/api/ai/chat` and `/api/ai/stream` routes, project-scoped session, `CHAT_CODE_SEARCH_TOOLS=true`), `investigation-coverage.test.ts` (19), 3 new end-to-end pipeline tests in `could-not-verify-pipeline.test.ts`, and 1 UI test in `gap-report.test.tsx`.

**Red on the unfixed code:** all 4 route tests fail on `main`, with the incident reproduced (`expected '{"response":{"content":"<tool_calls>\…' not to contain 'tool_calls'`, and `expected [] to deeply equal [ …(3) ]`, i.e. no tool ran). The two #19 pipeline assertions fail on `main` with `starved: false`.

**Mutations applied one at a time; every one turned the suite red:**

| # | Mutation | Red |
|---|---|---|
| M1 | multi-call result disabled (`>= 2` → `>= 99`) | 17 |
| M2 | scanner ignores string literals | 2 |
| M2b | scanner ignores escapes inside strings | 1 |
| M3 | "findings anywhere" safety rule removed | 1 |
| M4 | only the first call of a batch executed | 5 (unit + route) |
| M5 | per-reply cap removed | 1 |
| M6 | only the last result returned to the model | 2 |
| M7 | post-loop guard reverted to `parseToolCall` | 4 (unit + route) |
| M8 | default final-answer validity reverted | 2 |
| M9 | `<tool_call` tag detection removed | 7 |
| M9b | tag detection flags any mention | 1 |
| M10 | `isJsonFinalAnswer` reverted | 1 |
| M11 | `classifyFinalAnswer` reverted | 1 |
| M12 | abort check between batch calls disabled | 1 |
| M13 | quadratic scanner (rescan after an unclosed opener) | 2 (timing) |
| M14 | search-starvation rule disabled | 6 |
| M15 | "most" → "at least half" | 2 |
| M16 | exhaustion exemption removed | 2 |
| M17 | orchestrator persists the verdict-side record | 2 (pipeline) |
| M18 | a verified sibling finding no longer rescues a requirement | 1 |
| M19 | merge drops the unverified count | 1 |
| UI | "far fewer searches" wording removed | 1 |

The first version of the string-literal test **did not fail under M2**: its in-string braces were balanced. It was rewritten with unbalanced braces and an escaped quote, and both M2 and M2b now fail it.

## Coverage (vitest v8, suites `src/lib/analysis`, `src/lib/ai`, AI route and agent-loop tests)

|---|---|---|
| `agent-loop.ts` | 98.01 → 98.39 | 91.28 → 92.50 |
| `retrieval-health.ts` | 100 → 100 | 92.78 → 93.49 |
| `agentic-degradation.ts` | 93.33 → 93.33 | 86.84 → 86.84 |
| `orchestrator.ts` (subset run; the full suite covers more) | 58.68 → 58.88 | 49.71 → 50.28 |

The new orchestrator lines are covered by the pipeline tests (M17 fails them).

## Gate (run locally in an isolated worktree)
- `pnpm lint`: exit 0, including the private-terms, NUL, licence and CLA checks.
- `pnpm typecheck`: exit 0.
- `pnpm test`: server 1102 files / 17,235 tests passed; ui 350 / 3,239; shared 36 / 698; copilot-svc, embeddings-svc and ui-kit all passed. **scripts** failed 2 tests on the first run with **timeouts** (`Test timed out in 5000ms`) in the tree-scanning identifier-gate tests, at load average 48 from concurrent sessions. It passed 54/54 on a re-run at load 4. Neither test touches this diff.
- `node scripts/verify-changelog-fragment.mjs`: pass. There are two fragments, one per issue.

## Security
- There are no dependency changes, so no audit was needed.
- Both new parsers walk untrusted model output in linear time (per-character loops, no regex), and ReDoS tests cover 7 adversarial shapes.
- The per-reply cap bounds how much tool work one reply can trigger.
- Tool execution stays scoped to the session's project through the unchanged `toolContext`.
- UI values render as React text, with no HTML sinks.
- There is no SQL and no shell.

## Not done here
- Qwen-style `<tool_call>{"name":…, "arguments":…}` bodies are still not *executed*: `toolCallFromValue` requires `tool`. The markup is now caught and never rendered.
- The protocol text still says "Call tools one at a time". It was left unchanged to keep the cached prompt prefix byte-stable.
- `analysis-export.ts`'s degraded warning does not yet mention `unverifiedRequirements`.


mgcronin (commented):
Author notes on the load-bearing choices (self-review; not an approval).

server/src/lib/analysis/agent-loop.ts
```diff
@@ -761,6 +770,185 @@ export function parseToolCall(
...
+      continue; // prose in braces, or a broken call — neither is a call
+    }
+    const items = Array.isArray(value) ? value : [value];
+    for (const item of items) {
+      if (isFindingsAnswer(item)) return null;
+      const call = toolCallFromValue(item, known);
+      if (call) scanned.push(call);
+    }
+  }
+  if (scanned.length >= 2) return scanned;
```
mgcronin: Precedence: a multi-call result only wins at 2+ calls. At 0 or 1 calls, the unchanged `parseToolCall` decides, so every single-call reply parses exactly as before, including the fenced and first-to-last-brace cases that the #1244 differential fuzz pins.

server/src/lib/analysis/agent-loop.ts
```diff
@@ -761,6 +770,185 @@ export function parseToolCall(
...
+}
+
+/**
+ * #15 — the most tool calls executed from ONE model reply. The loop's turn cap
+ * bounds how many REPLIES a run gets, not how many calls a reply may ask for, and
+ * the reply is untrusted model output — without a cap one reply could request
+ * hundreds of searches. Calls past the cap are not run; the model is told which
+ * ones were dropped so it can ask again next turn.
+ */
+export const MAX_TOOL_CALLS_PER_REPLY = 8;
```
mgcronin, line 790: The reply is untrusted, and the turn cap limits replies, not calls per reply. 8 is well above what any observed model asks for in one turn (the DeepSeek incident asked for 3). Calls past the cap are reported back to the model, not silently dropped.

server/src/lib/analysis/agent-loop.ts
```diff
@@ -1284,7 +1495,7 @@ export async function runAgentLoop(
   // model was still emitting a tool call, `finalResponse` is raw `{"tool":...}`
   // protocol JSON. Never hand that to a caller: substitute a safe,
   // human-readable fallback. Pure post-loop sanitization — no model call.
-  if (parseToolCall(finalResponse) !== null) {
+  if (isToolCallReply(finalResponse)) {
```
mgcronin: The guard now also catches markup that does not *parse* (a truncated call, or a broken `<tool_calls>` wrapper), which is the #15 AC "never rendered to the user". A tag that is only mentioned in prose is not flagged; see `hasToolCallTagWrappingJson` and its negative test.

server/src/lib/analysis/orchestrator.ts
```diff
@@ -2450,6 +2451,27 @@ export class AnalysisOrchestrator {
...
+          if (reportedHealth.degraded && !passHealth.degraded) {
+            log.warn("Agentic code pass investigated too little to report as healthy", {
+              analysisId: input.analysisId,
+              requirementCount: passRequirements.length,
+              codeRetrievalCalls: reportedHealth.totalCalls,
+              unverifiedRequirements: reportedHealth.unverifiedRequirements ?? 0,
+              starved: reportedHealth.starved,
+            });
+          }
+          passHealths.push(reportedHealth);
```
mgcronin, line 2474: Only the report-side record is pushed. `passHealth`, which the verdict gate and `absenceIsConfirmableForClaim` read above, is untouched, so #19 cannot change a verdict. M17 (pushing `passHealth` instead) turns the #19 pipeline tests red.

server/src/lib/analysis/retrieval-health.ts
```diff
@@ -588,11 +588,113 @@ export function mergeRetrievalHealth(
...
+): AnalysisRetrievalHealth {
+  const unverified = Math.min(
+    Math.max(0, Math.floor(input.unverifiedRequirements)),
+    health.requirementCount,
+  );
+  const assessed: AnalysisRetrievalHealth = {
+    ...health,
+    ...(unverified > 0 ? { unverifiedRequirements: unverified } : {}),
+  };
+  if (health.exhausted === true || health.requirementCount === 0) return assessed;
```
mgcronin: The exhaustion exemption respects #1236, and the existing exhausted-pass pipeline test asserts `starved: false, degraded: false`. If reviewers want budget-cut runs to raise the warning too, this is the line to change, and that test's assertion would have to change with it.

server/src/lib/analysis/retrieval-health.ts
```diff
@@ -588,11 +588,113 @@ export function mergeRetrievalHealth(
...
+export function countUnverifiedRequirements(
+  findings: readonly VerdictBearingFinding[],
+  requirementIds: Iterable<string>,
+): number {
+  const inPass = new Set(requirementIds);
+  const verified = new Map<string, boolean>();
+  for (const finding of findings) {
+    const id = finding.requirementId;
+    if (!id || !inPass.has(id) || finding.verdict == null) continue;
+    const thisVerified = finding.verdict !== "could-not-verify";
```
mgcronin: **BLOCKING (Medium, high confidence): the unverified count disagrees with the verdict the page shows, so #19's 'most requirements could-not-verify ⇒ degraded' is only partly met.**

`deriveRequirementVerdict` (requirement-verdict.ts:459) rolls a requirement up to `could-not-verify` when it has **no** linked code finding (rule 4, the budget-starvation rule), and it ranks `could-not-verify` **above** `implemented`. This function does the opposite on both points: an unreported requirement is skipped, and a single non-CNV sibling marks the requirement verified.

I checked this with a scratch script against this branch:
- 16 requirements, 3 findings (`implemented`) for R0 to R2, nothing for the other 13: `countUnverifiedRequirements` returns **0**. With `totalCalls: 5` the assessed record is `starved:false, degraded:false`, while the requirements table shows 13/16 could-not-verify.
- Every requirement with one `implemented` and one `could-not-verify` finding: count **0**, rollup CNV for all 16.
- Worst case: 4 searches for 16 requirements (4 is not < 16 × 0.25) and an agent that reports on none: healthy, 0/16 verified.

The #19 incident was caught only because the starvation floor also fired. The `unverifiedRequirements` count also feeds the UI's `N of M requirements could not be verified`, which would contradict the table.

Suggested fix: count a requirement in the pass that has no verdict-bearing finding as unverified, and make CNV win over `implemented`, matching `deriveRequirementVerdict`. Better, derive it from the same rollup. Add tests for both shapes.

server/src/lib/analysis/retrieval-health.ts
mgcronin: Advisory (Medium-low, medium confidence): the exhaustion exemption skips **both** rules. #1236 was about a budget cut-off retitling findings, which is verdict-side, and this function is report-side only. So the `mostlyUnverified` rule is exempted without a stated reason.

Probe: `exhausted: true, totalCalls: 1, requirementCount: 16, unverifiedRequirements: 16` returns `degraded: false`. When the final-answer retry succeeds (`hasFinalAnswer` true), `markCodeAgentDegraded` is not called either, so such a run raises **no** banner even though nothing was verified.

Consider keeping the exemption for `searchStarved` only, and still setting `degraded` when most requirements are unverified. The existing #1236 test asserts on a pass whose findings were verified, so check whether it would even change.

server/src/lib/analysis/agent-loop.ts
mgcronin: **Medium (high confidence), regression, advisory: legitimate chat answers are now executed as tool calls.** The scan accepts any object with a `tool` string, whether or not the tool is registered. Before this PR, both of these returned `null` from `parseToolCall` and were treated as the answer:

```
Your linters:
```json
[{"tool":"eslint","version":"9"},{"tool":"prettier","version":"3"}]
```
```
→ now `[{tool:"eslint"},{tool:"prettier"}]`

`The protocol looks like {"tool":"search_code","args":{}} and then {"tool":"read_file","args":{}}.` → now two calls.

In chat, the answer is discarded, the model receives `Unknown tool` errors, and if it repeats the answer the turn cap replaces it with the fallback message. Analysis is protected by the `findings` rule; chat is not.

Suggested fix: in the **multi-call** branch only, accept a scanned call only when `knownTools` is supplied and the name is registered, or only when the text outside the spans is whitespace, fences or `<tool_call(s)>` tags. Keep the single-call path unchanged for the #1244 fuzz parity. Add a negative test for each shape.

File: server/src/lib/analysis/agent-loop.ts [hunk 7]
```diff
@@ -761,6 +770,185 @@ export function parseToolCall(
...
+    if (inString) {
+      if (ch === "\\") i++;
+      else if (ch === '"') inString = false;
+      continue;
+    }
+    if (ch === "{" || ch === "[") {
+      if (depth === 0) start = i;
+      depth++;
+    } else if (ch === "}" || ch === "]") {
+      if (depth === 0) continue;
```
mgcronin, line 848: Low (high confidence): an unmatched opener in prose before the calls swallows the rest of the reply as one unterminated span, which the comment above documents. That loses every call, and because the reply does not start with `{"tool"` and has no tag, `looksLikeToolCallMarkup` is false too. So the raw tool JSON is rendered as the answer.

Probe: `Let me check { quickly.\n{"tool":"a","args":{}}\n{"tool":"b","args":{}}` gives `parseToolCalls → null` and `isToolCallReply → false`. That is an edge case of #15's 'never rendered' criterion. A cheap mitigation that stays linear: when a span never closes, restart the scan once at the next `{` after `start`, or treat any `{"tool"` appearing after a newline as markup in `looksLikeToolCallMarkup`.

server/src/lib/analysis/orchestrator.ts
mgcronin, line 2474: Good call: pushing `reportedHealth` while the gate keeps reading `passHealth` keeps #19 off the verdict path. I confirmed nothing downstream re-runs `absenceIsConfirmable` on the persisted record (the only other call, `:2819`, is on the single-shot `groundedHealth`).

ui/src/components/analysis/gap-report.tsx
```diff
@@ -213,7 +213,13 @@ function RetrievalPanel({ retrieval }: { retrieval: AnalysisRetrievalHealth }):
         <span className="font-normal">
           ({retrieval.successfulSearches} of {retrieval.totalCalls} retrieval calls returned results
           {/* #1236 — budget cut-off is `exhausted`, not `starved`. */}
-          {retrieval.exhausted ? "; the investigation was cut short by its budget" : ""})
+          {retrieval.exhausted ? "; the investigation was cut short by its budget" : ""}
+          {/* #19 — a run can clear the search threshold and still check almost nothing. */}
+          {retrieval.starved ? "; far fewer code searches than requirements" : ""}
+          {retrieval.unverifiedRequirements
```
mgcronin, line 224: nit (Low): when `degraded` comes only from `mostlyUnverified` (searches worked, the agent still verified little), the panel headline at `:211` still reads 'Code search returned little usable evidence — "not found" results are unreliable'. That misattributes the cause. Consider a headline that depends on which signal fired.

mgcronin (commented):
Recommendation: do not merge — 1 blocking finding (plus the repo-wide `api` image-budget failure, which is not this PR's fault but still blocks a merge under CLAUDE.md)

This is an automated analysis published as a COMMENT, not a human sign-off. The reviewing gh identity (mgcronin) is also the PR author, so GitHub cannot record an approval or change request from it. A human owns the merge decision.

## Code Review Summary — PR #37 (closes #15, #19)

### Requirements: PARTIAL
**#15**
- [x] Consecutive objects, JSON array, and `<tool_calls>` wrappers (flat and nested DeepSeek form) parse into an ordered list
- [x] Calls run in order; every result goes back on the next turn in one message; a single call's message is unchanged
- [~] Unparseable tool markup is never rendered: true for truncated calls and broken wrappers. Not true when a stray `{`/`[` in prose comes before the calls (see the inline comment, Low)
- [x] Linear time: one bracket scan, one `JSON.parse` per disjoint span, index-walk tag detection, no regex
- [x] Tests for each encoding, plus chat and stream route tests
- New regression (Medium): legitimate chat answers that contain two or more `{"tool": ...}` objects, or a JSON array of them, are now executed as tool calls (inline)

**#19**
- [x] One search for 16 requirements gives `starved: true, degraded: true` and raises the banner
- [ ] **"Most requirements could-not-verify ⇒ degraded" is only partly met (BLOCKING).** `countUnverifiedRequirements` counts differently from the requirement verdicts the page shows (`deriveRequirementVerdict`). A run with 4 searches for 16 requirements whose agent reports on none of them is recorded healthy, while the page shows 16/16 could-not-verify (inline)
- [~] Budget-exhausted passes skip the unverified-share rule as well as the search floor (inline, advisory)

### Security: CLEAN (no findings)
Model output is untrusted input. It is parsed in linear time. Calls per reply are capped at 8, and the model is told which were skipped. Tools stay scoped to the session's `projectId` through `toolContext`. No SQL, shell, HTML or new dependencies. The false-positive execution above only runs read-only, project-scoped tools, so it is a correctness issue, not a security one.

### Code quality: GOOD
The design is sound: `toolCallFromValue` is shared so single and multi calls follow the same rules, the "findings anywhere" safety rule is kept, and `assessInvestigationCoverage` is report-side only, with the verdict gate still reading `passHealth`. I checked that no consumer re-evaluates `absenceIsConfirmable` on the persisted record, so overloading `starved` does not reach verdicts.

### Performance: CLEAN

### Tests: SUFFICIENT for what they cover
- **Verified the fail-before claims myself.** I reverted `agent-loop.ts` and `agentic-degradation.ts` to origin/main: all 4 route tests in `tests/ai-routes-multi-tool-call.test.ts` fail. I then reverted `orchestrator.ts` and renamed `assessInvestigationCoverage` away: both #19 pipeline tests fail (`expected false to be true`). With the PR restored, 7 files and 150 tests pass. Worktree restored clean (`git status` empty).
- Missing negatives: a JSON-array or prose answer with `"tool"` keys; a requirement with no finding, or with mixed verdicts, counted as unverified.
- Coverage on touched files is above 80% except `orchestrator.ts` (58.9%). That shortfall is pre-existing, and the new lines are exercised by the pipeline tests.

### Documentation: UP TO DATE
Two fragments (`15-…`, `19-…`) follow the `.changes/unreleased/<issue>-<slug>.md` convention, and the `changelog` check passes. The ARCHITECTURE.md paragraph is accurate. Deferred, and acceptable: the #744 export does not show `unverifiedRequirements`, and the protocol text at `agent-loop.ts:586` and `:609` still says "Call tools one at a time", kept for prompt-cache stability. Worth a follow-up issue, because the prompt now contradicts what the loop supports.

### CI
Every check passes except `api` and `api-outcome`. Both fail at **Verify image size budget**. Lint, Typecheck and Test all passed in that job (from `gh run view 35670268535 --json jobs`). This is a real failure, not a supersede cancel, and PR #36 fails the same way. It is tracked by #34 with fix PR #38. It is not caused by this PR, but CLAUDE.md forbids merging with genuinely failing checks, so re-run `api` once #38 lands. `e2e` and `windows` were skipped.

### Scope of this review
I read all 14 changed files and the surrounding orchestrator, rollup and chat-runtime code. I ran the new and adjacent server tests, did the revert-and-fail check, ran eslint on the changed files (clean) and ran `tsc --noEmit` on server (clean). I probed edge cases with a scratch tsx script. I did not rerun the full `pnpm test`: CI's `api` Test step and the `ui` job cover it. I did not run the live app.

mgcronin:
## Changes after review 5273128961 (commit 9396b40)

**1. Blocking. `countUnverifiedRequirements` now matches the page.** Each requirement in the pass is rolled up with `deriveRequirementVerdict`, the same function the page uses. A requirement with no finding counts as could-not-verify, could-not-verify beats implemented, and gap-confirmed beats both. The reviewer's three cases now give:
- 3 implemented and 13 with no finding: 13 unverified, degraded.
- implemented plus could-not-verify on every requirement: 16 unverified.
- 4 searches for 16 requirements and no findings: 16 unverified, degraded.

A new test builds 40 requirements with a mix of verdicts and checks that the count equals the number of `deriveRequirementVerdict` results that are could-not-verify.

**Behaviour change to review.** Four existing tests assumed a run was healthy when its only requirement showed "could not verify" on the page:
- the #1236 exhausted-run pipeline test;
- the #773 title-injection test;
- two orchestrator "clean run" tests, whose fixture finding had no `verdict`.

Under the page's rules those runs are mostly unverified, so they now report `degraded`. `starved` is still false in each. The pipeline tests now assert `degraded` and `unverifiedRequirements`. The clean-run fixtures now return an `implemented` verdict that cites the code their search returned. The #773 scale test's "gap-heavy run is not degraded" case now runs at N=5 rather than N=20: at 20, 16 filler requirements have no finding, so the run is correctly degraded, and that case is about gaps, not coverage. In production, this means a run whose findings leave out `verdict` will now raise the banner, because the page already shows those requirements as unverified.

**2. Advisory, taken.** A reply with several `{"tool":…}` objects now counts as tool calls only if at least one names a registered tool, or the reply uses a `<tool_call>`/`<tool_calls>` tag. The checks after the loop (`turnsExhausted`, `isValidFinalAnswer`, the fallback guard) now use the same tool set as the loop. So a chat answer that is a JSON array, or prose quoting `eslint`/`prettier` tool objects, is returned as the answer. A single unknown-tool object is handled as before and still gets the repair error.

**3. Advisory, taken.** The "most unverified" rule now also applies to runs that used up their turn or token budget. Those runs are still never marked `starved`.

**4. Advisory, taken.** When a `{` or `[` in prose never closes, the scan starts again just after it, at most 4 times, so the calls that follow are still found. Each rescan is one linear pass, and two ReDoS shapes were added to the linear-time tests.

**5. Nit, taken.** When the warning comes only from the "most unverified" rule, the panel headline now says "Most requirements could not be verified against the code" instead of blaming code search.

**6. Informational.**
- The "Call tools one at a time" prompt text is tracked in #40.
- The export still does not show `unverifiedRequirements`. It stays deferred, out of scope here.
- `orchestrator.ts` is not changed in this commit.

**Tests fail without the fix.** Each mutation was applied, the tests run, and the code restored:
- old count (skipping requirements with no finding): 7 new tests red.
- dropping requirements with no finding from the new count: 5 red.
- putting back the exhaustion exemption for "most unverified": 2 red.
- exempting exhausted runs from the starvation rule removed: 2 red.
- gate on the registered-tool check forced true: 4 red.
- the tag exception disabled: 1 red.
- `isToolCallReply` "parses but none registered" line removed: 1 red.
- `turnsExhausted` without the registered tools: 2 red.
- `MAX_UNTERMINATED_RESCANS = 0`: 2 red.
- UI headline forced to the old text: 1 red.

**Gates (run locally in the worktree).**
- `pnpm lint`: exit 0.
- `pnpm typecheck`: exit 0.
- `pnpm test`: server 17255 passed, ui 3240 passed, scripts 2256 passed, shared 698 passed, ui-kit, copilot-svc and embeddings-svc all green, each package run on its own. In the full parallel fan-out, `scripts/lib/gate-cwd-anchoring.test.mjs` timed out at 5184 ms against its 5000 ms `testTimeout`. It runs the private-terms gate twice, about 2 s each, and it passes when run alone. This PR does not touch `scripts/`; it is the contention pattern described in #1379.
- Coverage on touched files: `agent-loop.ts` 97.6% lines, `retrieval-health.ts` 100% lines, `gap-report.tsx` 100% lines.

mgcronin:
## Independent re-verification of 9396b40

- **The reviewer's scenarios, re-run in a scratch test at 9396b40 (not committed):** 3 implemented and 13 with no finding gives 13 unverified and `degraded`. Implemented plus could-not-verify on every requirement gives 16. 4 searches, 16 requirements and no findings gives 16 and `degraded`. Exhausted with 16 of 16 unverified gives `degraded:true` and `starved:false`.
- **Mutations re-applied independently, each restored afterwards:**
  - revert `retrieval-health.ts` to 5d0fb1d: 9 red;
  - revert `agent-loop.ts` to 5d0fb1d: 6 red;
  - `MAX_UNTERMINATED_RESCANS = 0`: 2 red;
  - registered-tool gate forced true: 4 red;
  - exhaustion exemption restored: 2 red;
  - requirements with no finding skipped in the count: 5 red;
  - UI headline forced to the old text: 1 red.
- **Gate, local, at 9396b40:** `pnpm lint`, `pnpm typecheck` and `pnpm test` all exit 0 (server 17255 passed, ui 3240, scripts 2256, shared 698).
- **Coverage:** `agent-loop.ts` has 98.5% lines and 90.6% branches. `retrieval-health.ts` has 100% lines and 93.5% branches.
- **CI:** `api` / `api-outcome` fail only at "Verify image size budget" (#34). Lint, Typecheck and Test pass inside that job. Re-run `api` after #38 lands. Every other check passes.

server/src/lib/analysis/agent-loop.ts
```diff
@@ -813,7 +1047,8 @@ export type FinalAnswerKind =
 export function classifyFinalAnswer(text: string): FinalAnswerKind {
   const trimmed = (text ?? "").trim();
   if (!trimmed) return "empty";
-  if (parseToolCall(trimmed) !== null) return "tool-call";
+  // #15 — a multi-call reply is a tool call too, not "valid-json" or prose.
+  if (parseToolCalls(trimmed) !== null) return "tool-call";
```
mgcronin, line 1051: Low, medium confidence: `classifyFinalAnswer` still calls `parseToolCalls(trimmed)` without the registered tools. Inside `runAgentLoop` (`:1510`), a retry answer that quotes an unregistered-tool JSON array is therefore classified `tool-call`, while `isValidFinalAnswer` treats the same text as an answer. The effect is limited to telemetry and repair triage, so this is not blocking. Consider an optional `knownTools` parameter that the loop passes.

server/src/lib/analysis/agentic-degradation.ts
```diff
@@ -71,7 +71,9 @@ const REASON_TEXT: Record<AgenticDegradationReason, string> = {
  * when it is NOT a tool call AND it contains a parseable JSON object.
  */
 export function isJsonFinalAnswer(text: string): boolean {
-  if (!text || parseToolCall(text) !== null) return false;
+  // #15 — `isToolCallReply`, not the single-call parser: a multi-call reply's
+  // first object would otherwise pass `extractJsonObject` below as an "answer".
+  if (!text || isToolCallReply(text)) return false;
```
mgcronin, line 76: Low: `isJsonFinalAnswer` calls `isToolCallReply(text)` without the registered tools, so it is stricter than the loop default. An analysis answer shaped like `[{"tool":…},{"tool":…}]` is rejected. That is harmless because a real analysis answer is a findings object, so this is not blocking; it is noted only for consistency with advisory 2.

mgcronin (commented):
Recommendation: merge once CI is green — no blocking code findings. It is not merge-ready yet: `api` / `api-outcome` fail at "Verify image size budget" (#34). That failure is not caused by this PR. Fix PR #38 is still OPEN; re-run `api` after it lands.

This is an automated re-review (cycle 2) of head 9396b40, published as a COMMENT. It is not a human sign-off, and gh user `mgcronin` is also the PR author.

### Previous blocking finding: fixed, checked by running it
`countUnverifiedRequirements` now rolls each requirement up with `deriveRequirementVerdict`. I ran a scratch tsx script against `retrieval-health.ts` (deleted afterwards; worktree clean) with 16 requirements:
- 3 implemented + 13 with no finding, 5 calls: 13 unverified, **degraded**. It was 0 and healthy before.
- implemented + could-not-verify on every requirement: 16 unverified, **degraded**. It was 0 before.
- no findings, 4 calls: 16 unverified, **degraded**. It was healthy before.
- budget exhausted, no findings: 16 unverified, **degraded**, not starved. This is advisory 3, fixed.
- controls: all implemented gives 0 and not degraded. 8 of 16 unverified sits exactly at the 0.5 share and is not degraded, as the strict `>` intends.

### Previous advisory findings
- 2 (unregistered tools in the multi-call scan): fixed. The scan counts only if a call names a registered tool or the reply has a tool tag. The loop (`:1350`), `turnsExhausted`, the final-answer check and the sanitiser all pass `registeredTools`.
- 4 (unclosed opener): fixed. There are at most 4 rescans, each starting past the previous opener, so parsing stays linear and spans are not duplicated.
- 5 (UI headline): fixed. The tests for it pass.
- 6: the prompt text is tracked in #40.

### New finding (low, not blocking)
Two call sites still parse without the registered tool list; see the inline comments. Both are in the analysis path only, where the findings-JSON rule already guards.

### Checked
- The touched server suites pass: 7 files, 239 tests (multi-tool-calls, investigation-coverage, could-not-verify-pipeline, analysis-orchestrator, agent-loop*). `gap-report.test.tsx` passes 19 tests.
- I did not re-run the full `pnpm lint && typecheck && test`. The author reports it green at 9396b40, and CI’s `api` job shows Lint, Typecheck and Test passing before the image-size step.
- CI: the only failing step is Verify image size budget. Everything else passes; `e2e` and `windows` were skipped.
- Security: no new findings.

PR #38, opened 2026-09-22
fix: Resolve #34, #3 — slim metis-server image, cache CI image builds

### LINKED ISSUE #3: Docker layer caching silently became a no-op on hosted runners — api builds cold every run
**Two Docker layer caches stopped working when CI moved to GitHub-hosted runners, because both were properties of the machine rather than of the workflow.**

Neither failed loudly. Both silently became no-ops, and the first migrated `api` run was still building at 30 minutes against a 45-minute timeout.

## The two dead caches

**1. `ci.yml`'s `api` job — an implicit daemon cache.**

The job uses plain `docker build` rather than a buildx container, deliberately: bootstrapping `moby/buildkit` intermittently hung and once stalled the job for 30+ minutes. The comment at `ci.yml:118` explains the trade — *"The daemon's layer cache still makes warm rebuilds fast"*.

That daemon was the self-hosted one, persistent across runs. A GitHub-hosted runner is a **fresh VM every time**, so there is no daemon to hold layers and every build is cold.

**2. `build-images.yml` — an explicit local-path cache.**

```yaml
cache-from: type=local,src=/tmp/buildkit-cache/${{ matrix.image.name }}
cache-to:   type=local,dest=/tmp/buildkit-cache/${{ matrix.image.name }},mode=min
```

`/tmp/buildkit-cache` persisted on the self-hosted box. On a hosted runner it is empty at the start of every run and discarded at the end — so `cache-to` writes into a directory nobody will ever read.

## Stopgap currently in place

`api`'s `timeout-minutes` is raised from 45 to 90 so the migration could land. **That is a stopgap and the comment says so.** A raised timeout hides a slow build rather than making it fast, and 90 minutes of cold rebuild on every pull request is a real tax on contribution.

## The fix

Use a cache backend that survives the VM. For buildx jobs that is:

```yaml
cache-from: type=gha
cache-to:   type=gha,mode=max
```

`build-images.yml` already uses `docker/build-push-action`, so that change is nearly mechanical.

`ci.yml`'s `api` job is the harder one, because it deliberately avoids buildx. Options, and they should be measured rather than assumed:

1. Move it to `docker/build-push-action` with the `gha` backend, accepting the BuildKit container it was avoiding — the hang that motivated avoiding it was on the self-hosted daemon and may not reproduce here.
2. Keep plain `docker build` and cache the expensive inputs instead (the pnpm store, the built `node_modules`), which does not cache layers but may recover most of the time.
3. Ask whether `api` needs to build all three images on every pull request at all. It is an image-size regression gate; `build-images.yml` is the authoritative multi-arch build.

## Acceptance criteria

- [ ] `api` completes well inside a sane timeout on a warm cache, measured before and after
- [ ] `build-images.yml` uses a cache that survives the runner, with the local-path form removed
- [ ] `api`'s `timeout-minutes` is lowered from 90 and the stopgap comment deleted
- [ ] A cold-cache run is measured too, so the first build after a cache eviction is a known quantity rather than a surprise

### LINKED ISSUE #34: metis-server image is 1,301 MB against the 350 MB budget — api fails on every pull request

mgcronin:
Closes #34
Closes #3

## Summary

**#34, image size.** `metis-server` goes from **1,301 MB to 821 MB** on amd64, as measured by CI (`docker image inspect`, decimal MB). The size gate now holds it to its own budget of **900 MB**, which is the measurement plus about 10% headroom. `metis-ui` stays on the general 350 MB (206.7 MB measured).

**#3, CI build cache.** `api` builds its four images with `docker/build-push-action` and a `type=gha` cache. That cache survives the hosted VM, which the old daemon cache and the `/tmp/buildkit-cache` local cache did not. `timeout-minutes` drops from 90 to 60, and the stopgap comment is gone.

## What changed

- `Dockerfile.server`: a reachability prune, `scripts/lib/prune-pnpm-store.mjs`, keeps only the store entries reachable from the runtime importers. It excludes the `prisma` CLI and the Copilot SDK's CLI binary. Pruning by reachability instead of by glob matters because a glob that stops matching fails silently, and that is how the Copilot CLI and Prisma 7's query compilers got back into the image.
- `Dockerfile.server` also drops tree-sitter C sources and native prebuilds (the server loads only the `.wasm`), other platforms' `prebuilds/`, better-sqlite3's SQLite sources, Oracle's JDBC jars, and npm/npx/yarn/corepack from the runtime stage.
- `scripts/lib/verify-image-size.mjs` and `scripts/verify-image-size.sh` add a per-image server budget (`MAX_SERVER_IMAGE_MB`, default 900). A larger server budget does not loosen the UI gate.
- `ci.yml` (`api`): buildx with a `type=gha` cache (`mode=max`, one scope per image), a step that frees runner disk before the builds, and a 60-minute timeout.
- `build-images.yml`: the local-path cache is replaced with `type=gha`, using the same scopes as `ci.yml`. The release-tag build only reads the cache (see below).
- `docs/OPERATIONS.md` → *Container Image Sizes* and *CI build cache (#3)*: the measured breakdown, the budget and why it is provisional, and the cold and warm timings.
- Tests: `prune-pnpm-store.test.ts`, `verify-image-size.test.ts` and `image-build-wiring.test.mjs`.

## Measurements

**Image size** (amd64, CI): server 1,301 → 821 MB (run 35673206512). The removal breakdown is in `docs/OPERATIONS.md`.

**`api` job timing** (hosted `ubuntu-latest`, from step timings via `gh api repos/openzigs/metis/actions/jobs/<id>`):

| Cache | Run (job) | server | ui | embeddings | sql-lineage | four builds | whole job |
|---|---|---|---|---|---|---|---|
| Before #3: stopgap timeout | first migrated run | | | | | | still building at 30 min (limit raised to 90) |
| Cold, empty cache | 35671126250 (106567592133) | 7m19s | 3m42s | 2m11s | 18s | 13m30s | 28m20s |
| Partly warm, server source changed | 35673206512 attempt 1 (106574086399) | 3m03s | 2m47s | 1m49s | 6s | 7m45s | 23m22s |
| Warm, re-run of the same commit | 35673206512 attempt 2 (106589624773) | 2m31s | 2m33s | 2m12s | 9s | 7m25s | 18m46s |

A warm cache roughly halves the image builds and puts the job at under 19 minutes, against a 60-minute timeout. Warm builds do not get down to seconds. Every dependency-install layer restores as `CACHED`, but the compile steps after the source `COPY` still re-ran in the warm run, and restoring and exporting the cache costs 15–60 s per image. `Test` alone takes 8–11 minutes of every run.

## The 900 MB budget is provisional

The budget is a measured ceiling, not a floor. Two of the largest remaining components do not work in this image today (#39). The Oracle Instant Client (~138 MB) fails `initOracleClient` with `DPI-1047`, and `@lancedb/vectordb` (~99 MB) fails to relocate on musl. Fixing #39 means re-measuring the budget. Other contributors that could still be cut: 3 copies of `@napi-rs/canvas` (~55 MB), unused `tesseract.js-core` WASM variants (~35 MB), `@azure/msal-browser` in a server image, and ~13 MB of better-sqlite3 build leftovers.

## Review follow-ups (review 5273531606)

Fixed in this PR:
- **B1**: the warm-cache row held placeholder tokens. It now has the measured figures above.
- **B2**: the budget rationale called 350 MB unreachable "without dropping features", counting components that do not load. It now says the 900 MB budget is provisional pending #39.
- **A4**: the tag-push build in `build-images.yml` no longer has `cache-to`. Only that tag could ever restore the cache it wrote, so it only used up the 10 GB repository limit. The wiring test now requires that step to be read-only.
- **nit**: the changelog fragment says 0.82 GB, not 0.9.
- Recorded the review's two lasting findings in `.claude/agent-memory/code-issue/`: a green `api` job does not mean the image runs, and a cache written from a tag ref cannot be read by any other ref.

Left as follow-ups (advisory):
- **A1**: the budget pin still allows a silent raise, for example to 1199. Pin it to `docs/OPERATIONS.md` and to the `900` hard-coded in `verify-image-size.sh`.
- **A2**: nothing tests the size-gate step itself, or the tree-sitter/prebuilds prune. Neutering the gate or setting `MAX_SERVER_IMAGE_MB: 99999` stays green.
- **A3**: PR runs write `mode=max` caches. This PR alone wrote 3.09 GB in 98 entries. Consider writing the cache only on `main` pushes, or using `mode=min` for the sidecars.
- **A5**: the disk-freeing step takes 45s–1m49s even when the runner already has 83 GB free. Gate it on a free-space threshold.
- **A6**: the prune leaves dangling symlinks in `.pnpm/node_modules`, and phantom dependencies are not treated as roots.
- **A7**: the legacy glob prune block still deletes the reachable `@prisma+debug` (#39).
- **A8, A9**: better-sqlite3 `build/Release/obj` and `sqlite3.a`, `tesseract.js-core` WASM variants, and `@azure/msal-browser`.
- Warm builds still re-run the compile steps. Worth finding out why before relying on the cache for more.

## Verification

- Local gate: `pnpm lint && pnpm typecheck && pnpm test` passed, as did `pnpm agents:verify` and `pnpm changelog:verify`.
- Mutation for A4: re-adding `cache-to` to the tag-push step turns both new wiring tests red.
- The reviewer built base and head images and imported every `server/dist` module in both. The failure sets are identical, so there is no runtime regression from the prune. The container does not boot on either image (`ERR_MODULE_NOT_FOUND jszip`), which is pre-existing and tracked in #39.


docs/OPERATIONS.md
```diff
@@ -405,10 +437,31 @@ pnpm verify:image-size
...
+hosted VM. Before building, both workflows delete preinstalled toolchains the
+build never uses (.NET, Android SDK, GHC, CodeQL) and the runner's cached
+Docker images.
+
+Measured on #38 (hosted `ubuntu-latest`):
+
+| | server | ui | embeddings | sql-lineage | whole `api` job |
+|---|---|---|---|---|---|
+| Cold cache | 7m19s | 3m42s | 2m11s | 18s | 28m18s |
+| Warm cache | WARM_SERVER | WARM_UI | WARM_EMB | WARM_SQL | WARM_JOB |
```
mgcronin: **Blocking (B1).** These are literal template tokens, and they would ship into `main` as-is. Issue #3's first acceptance criterion is "`api` completes well inside a sane timeout on a warm cache, measured before and after". Run 35673206512 already gives a partially warm figure (server 3m03s, ui 2m47s, embeddings 1m49s, sql-lineage 6s, job 23m22s), though the server build re-ran `COPY server` because the source changed. Record a truly warm run, for example a re-run of the same SHA. Also finalise the PR body, which still reads "Work in progress".

mgcronin (commented):
Recommendation: do not merge yet: 2 blocking findings, both small documentation fixes. The engineering is sound. I found no runtime regression, and the size gate can still fail.

*This is an automated review, not a human sign-off. The reviewing `gh` identity is also the PR author, so GitHub cannot record an approval or a change request from it. A human decides whether to merge.*

## What I verified

**Runtime: no regression (checked by building and running both images).** I built `Dockerfile.server` from `main` (45b125c) and from this branch (9b7012d) on an arm64 host, then ran the same checks on both:
- Imported all 917 non-CLI modules under `server/dist`. Both images give the same failure set. Once I stubbed the two packages that are already missing on `main` (`jszip`, `@prisma/debug`, both covered by #39), both images import every module except a `__tests__` helper.
- Diffed the `.pnpm` store. The prune removes 58 entries: the prisma CLI tree, the copilot musl binary, react/radix/chart.js, pglite and a few more. None of them is imported by `server/src` or `packages/shared/src`. A static scan of every kept package for `require`/`import` of a removed name found only `cytoscape/playwright.config.js` → `dotenv`, which is not a runtime path.
- Native modules (better-sqlite3, argon2, vectordb, oracledb thick, tree-sitter WASM init, pdf-parse) behave the same in both images. Several are broken on both, which matches #39.

The container boots with neither image: `ERR_MODULE_NOT_FOUND jszip`. That is already the case on `main` and #39 tracks it. It also means **nothing in CI would catch a prune that deletes something the runtime needs.** That affects how much weight the size gate should carry (see advisory A2).

**Size gate can still fail.** `api` run 35673206512 shows `metis-server 821 / 900`, `metis-ui 206.7 / 350`. I mutated the gate and the prune in a scratch copy of the tree: ignoring the per-image budget, never assigning it, an Infinity default, `isOverBudget` always false, the prune ignoring `--exclude`, skipping the transitive walk, never deleting, and dropping `--exclude prisma` from the Dockerfile. **All of those go red.** The survivors are listed below.

**Cache (#3).** `type=gha` is wired and working: `importing cache manifest from gha:…` appears for all four images, and `timeout-minutes` dropped from 90 to 60 with the stopgap comment removed. The cold job took 28m18s and the latest run 23m22s.

**Security.** No findings. `prune-pnpm-store.mjs` never shells out and only deletes direct children of the realpath'd store. The disk-freeing step removes fixed paths. `github.head_ref` appears only in `if:` expressions. The new actions are pinned by SHA. There is no `pull_request_target`.

## Blocking
1. **B1: unfinished record.** `docs/OPERATIONS.md:461` would ship literal `WARM_SERVER | WARM_UI | WARM_EMB | WARM_SQL | WARM_JOB` tokens, and the PR body is still a "Work in progress" stub. Issue #3's first acceptance criterion needs the warm figure recorded.
2. **B2: the 900 MB rationale is contradicted by #39.** The budget itself is honest: 821 MB measured, about 10% headroom, and the removal table matches what I measured. But the doc justifies 350 as unreachable "without dropping features", counting Oracle Instant Client (~138) and LanceDB (~99). #39, filed from this same work, shows neither of those loads in this image. Mark the budget as provisional and pending #39, so it is not later cited as a floor.

## Advisory
- A1: The budget pin allows a silent raise to 1199. That mutant survives.
- A2: Nothing tests the size-gate step itself, or the tree-sitter/prebuilds prune. Neutering the gate step, or adding `MAX_SERVER_IMAGE_MB: 99999`, stays green.
- A3: GHA cache footprint. This PR alone wrote 3.09 GB in 98 entries against the 10 GB repository cap.
- A4: `build-images.yml` tag-push `cache-to` writes caches that no other ref can read.
- A5: The free-disk step costs 1.5–2 minutes on a runner that already had 83 GB free.
- A6: The prune leaves 56 new dangling symlinks in the hidden hoist dir (`.pnpm/node_modules`), and phantom deps are not treated as roots.
- A7: The legacy glob block (unchanged lines 136–152) still deletes reachable `@prisma+debug`, which contradicts the new reachability prune. This is pre-existing; see #39.
- A8: better-sqlite3 `build/Release/obj` (10 MB) and `sqlite3.a` survive the prune.
- A9: Remaining cuts that this PR did not list: `tesseract.js-core` ships 12 WASM variants (~44 MB, of which roughly 2 are used), and `@azure/msal-browser` is in the server image.
- nit: the changelog says "about 0.9 GB"; the measured size is 0.82.

Good work: the reachability prune is the right structural fix for "a glob that stopped matching fails silently", and the wiring tests caught my `--exclude prisma` mutation. Filing #39 rather than burying the boot failure was the right call.

## Scope of this review
I read all 11 changed files in full, plus issues #34, #3 and #39. I built both images locally on arm64 and did not build amd64. The amd64-only branch is the `linux-x64*` prebuilds filter, which mirrors the arm64 branch I exercised. I ran the three new or changed test files (59 pass) plus 16 mutations. CI is fully green: `api`, `api-outcome`, `postgres-adapter` and the rest pass, and `e2e`/`windows` are skipped by design.

docs/OPERATIONS.md
```diff
@@ -366,33 +366,65 @@ by combining three changes:
...
+| Node.js 20 runtime (`/usr/local`) | ~104 | The `node` binary alone is 97 |
+| `@lancedb/vectordb-linux-x64-gnu` | ~99 | Single native library |
+| `@napi-rs/canvas` ×3 (pdf-parse and two `pdfjs-dist` versions each pin their own) | ~88 | Native, musl |
+| `tesseract.js-core` | ~44 | OCR WASM, via `officeparser` |
+| `@prisma/client` | ~41 | After the prunes |
+| mermaid, `@kubernetes/client-node`, playwright-core, cytoscape, pdfjs-dist ×2 | ~92 | |
+| Everything else in `node_modules` | ~212 | ~750 store entries |
+| Compiled server (`server/dist`) + Prisma schema | ~29 | |
+
+The first six rows are over 500 MB before any application JavaScript, so 350 MB is
```
mgcronin: **Blocking (B2).** The budget number is honest: 821 MB measured plus about 10% headroom. I reproduced the removal table (arm64: `du -sx /` goes from 1239 MB to 814 MB). The *rationale* is the problem. It says 350 is unreachable "without dropping features" and counts Oracle Instant Client (~138) and LanceDB (~99) as features. #39, found while doing this work, shows `oracledb.initOracleClient` failing with DPI-1047 (libnnz.so not on the loader path) and `vectordb` failing to relocate on musl. I reproduced both on main and on this branch. So about 237 MB of the "irreducible" floor does not work in this image today.

Suggest adding a sentence saying the 900 MB budget is provisional and pending #39. Fixing #39 might mean a glibc base (bigger), or dropping the Instant Client (138 MB smaller). Either way the budget needs re-measuring then. Without that note, a later reviewer will cite this paragraph as the floor.

docs/OPERATIONS.md
```diff
@@ -366,33 +366,65 @@ by combining three changes:
...
-**Hitting the original ≤250 MB target requires one of:**
-- Externalising LanceDB to its own sidecar (saves ~87 MB) — or replacing it
-  with `pgvector` riding on the existing Postgres
-- Reverting to `pdf-parse@1`, which doesn't depend on `@napi-rs/canvas`
-  (saves ~24 MB) — at the cost of weaker PDF support
-- Switching the runtime to `gcr.io/distroless/nodejs20-debian12` — has not
-  been measured but expected to net <30 MB
-
-Tracked as follow-up work (see PR description for issue #145).
+| Oracle Instant Client (basiclite 23.9, `.so` only) | ~138 | Thick mode for pre-12c password verifiers |
```
mgcronin: Supports B2: this row's "Thick mode for pre-12c password verifiers" is currently aspirational. In both images, `initOracleClient({ libDir: '/opt/oracle/instantclient' })` fails with `DPI-1047 … Error loading shared library libnnz.so`, even though the file exists at `/opt/oracle/instantclient/libnnz.so`. That makes it the largest item in the table and non-functional as shipped. It is the strongest candidate for the build-arg you list under "Further cuts".

scripts/lib/verify-image-size.test.ts
```diff
@@ -185,6 +208,25 @@ describe("parseHumanSizeToBytes", () => {
...
+    expect(resolveImageBudget(server, 350, {})).toBe(1000);
+    expect(resolveImageBudget(server, 350, { MAX_SERVER_IMAGE_MB: "junk" })).toBe(1000);
+    expect(resolveImageBudget(server, 350, { MAX_SERVER_IMAGE_MB: "  " })).toBe(1000);
+  });
+  it("an image with no budget of its own takes the general budget", () => {
+    expect(resolveImageBudget({}, 350, { MAX_SERVER_IMAGE_MB: "1200" })).toBe(350);
+  });
+  it("the shipped server default is a real number below the old arm64 allowance", () => {
+    expect(DEFAULT_MAX_SERVER_IMAGE_MB).toBeGreaterThan(DEFAULT_MAX_IMAGE_MB);
+    expect(DEFAULT_MAX_SERVER_IMAGE_MB).toBeLessThan(1200);
```
mgcronin, line 226: **Advisory (A1): this pin cannot catch the thing issue #34 warned about.** I mutated `DEFAULT_MAX_SERVER_IMAGE_MB` from 900 to 1199 and all 59 tests stayed green. "Do not simply raise the number to make CI pass" is exactly that edit. Pin it with `expect(DEFAULT_MAX_SERVER_IMAGE_MB).toBe(900)`, or better, parse the `≤ 900 MB` figure out of `docs/OPERATIONS.md` and assert equality. Then a raise has to touch the doc's rationale, and a reviewer sees the rationale change. Also, `scripts/verify-image-size.sh:37` hard-codes a second `900`; one line asserting the two agree would stop drift.

scripts/lib/image-build-wiring.test.mjs
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it("frees runner disk before the first image build", () => {
+    const free = api.indexOf("- name: Free runner disk for image builds");
+    const firstBuild = api.indexOf("- name: Build metis-server");
+    expect(free).toBeGreaterThan(-1);
+    expect(free).toBeLessThan(firstBuild);
+  });
+
+  it("dropped the 90-minute stopgap timeout", () => {
```
mgcronin, line 109: **Advisory (A2): the step this PR rewired most is the one it doesn't test.** In a scratch copy I replaced `run: node scripts/lib/verify-image-size.mjs --no-build` with `run: echo skip`, and separately added `MAX_SERVER_IMAGE_MB: "99999"` to the gate step's env. **Both stayed green.** Either change turns the size gate into one that cannot fail. Suggest an arm asserting that `Verify image size budget` exists, runs `verify-image-size.mjs`, comes after every `Build metis-*` step, and sets no `MAX_SERVER_IMAGE_MB`.

The Dockerfile prune is also unguarded. Changing the tree-sitter block to `rm -rf "$D"` (deleting the whole grammar packages, WASM included), or making the prebuilds filter keep nothing, stays green. The image still builds, and since CI never runs the image (#39), nothing downstream notices. Until #39's `/healthz` smoke lands, one assertion that the tree-sitter loop never deletes `*.wasm` would be worth having.

.github/workflows/ci.yml
```diff
@@ -158,26 +176,72 @@ jobs:
...
+        uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
+        with:
+          context: .
+          file: Dockerfile.server
+          tags: metis-server:ci-${{ github.run_id }}
+          load: true
+          provenance: false
+          sbom: false
+          cache-from: type=gha,scope=metis-server
+          cache-to: type=gha,scope=metis-server,mode=max
```
mgcronin, line 193: **Advisory (A3): cache footprint, measured.** `gh api repos/openzigs/metis/actions/caches` shows `refs/pull/38/merge` holding **3.09 GB across 98 entries** after two runs, against the default 10 GB repository cap. Caches are ref-scoped, so each open PR that touches Docker inputs writes its own copy. With about three such PRs open, LRU eviction starts removing `main`'s entries, which are the ones every PR reads, and #3's cold-build problem comes back intermittently. Options: write the cache only on `push` to `main` (PRs `cache-from` only), or use `mode=max` for `metis-server` alone and `mode=min` for the sidecars.

.github/workflows/build-images.yml
```diff
@@ -182,8 +197,8 @@ jobs:
           # '*.dockerbuild' artifact adding ~60 KB per image per push.
           provenance: true
           sbom: false
-          cache-from: type=local,src=/tmp/buildkit-cache/${{ matrix.image.name }}
-          cache-to: type=local,dest=/tmp/buildkit-cache/${{ matrix.image.name }},mode=min
+          cache-from: type=gha,scope=metis-${{ matrix.image.name }}
+          cache-to: type=gha,scope=metis-${{ matrix.image.name }},mode=max
```
mgcronin: **Advisory (A4).** This workflow runs only on tag pushes. A cache written from a tag ref can be restored only by that same tag ref, so no later run reads these entries: not `main`, not the next tag. They still count toward the 10 GB cap, as multi-arch `mode=max` writes, and push out `main`'s. Reading works (a tag run can restore `main`'s cache), so keep `cache-from` and drop `cache-to` here. The comment at line 179 ("either warms the other's amd64 layers") is true only in the `main` → tag direction.

.github/workflows/ci.yml
```diff
@@ -118,11 +111,32 @@ jobs:
       - name: Test
         run: pnpm test
 
+      # #3 — a run on main ran OUT OF DISK building these ("Free space left:
+      # 55 MB"). Drop the preinstalled toolchains this job never uses (.NET,
+      # Android SDK, GHC, CodeQL bundles, the runner's cached Docker images)
+      # before building. Measured on #38: 83 GB -> 110 GB free, in 1m21s.
```
mgcronin, line 117: **Advisory (A5).** The same log shows `/dev/root 145G 63G 83G` *before* this step. The runner already had 83 GB free, and the step took 1m49s on the latest run (the comment says 1m21s). The 55 MB out-of-disk in #3 came from a runner with about 14 GB free. Consider gating the `rm -rf` on `df --output=avail -BG / | tail -1` being below a threshold (say 30G), so it costs nothing when it isn't needed.

scripts/lib/prune-pnpm-store.mjs
```diff
@@ -0,0 +1,185 @@
...
+    visit(listLinks(path.join(storeReal, key, "node_modules")));
+  }
+
+  /** @type {string[]} */
+  const kept = [];
+  /** @type {string[]} */
+  const removed = [];
+  for (const e of fs.readdirSync(storeReal, { withFileTypes: true })) {
+    // `.pnpm/node_modules` is pnpm's hidden hoist dir; `lock.yaml` is metadata.
+    if (!e.isDirectory() || e.name === "node_modules") continue;
```
mgcronin, line 114: **Advisory (A6).** pnpm's hidden hoist dir `.pnpm/node_modules` holds a symlink to *every* store entry, which is how a package resolves a dependency it forgot to declare (a phantom dep). Skipping it as a root is the right call, but it has two consequences:
1. After the prune, the built image has **56 new dangling links** there (`dotenv`, `signal-exit`, `chokidar`, `zod@4`, `react`, …). I diffed the dangling symlinks: 144 on main, 186 here.
2. If a kept package ever phantom-requires a pruned name, it fails at runtime and not at build time.

I scanned every kept package for `require`/`import` of a removed name. The only hit was `cytoscape/playwright.config.js` → `dotenv`, which is not a runtime path, so nothing breaks today. Suggest deleting dangling links under `.pnpm/node_modules` in the same pass and adding a sentence to the header doc naming the phantom-dep limitation.

Dockerfile.server
```diff
@@ -156,6 +157,48 @@ RUN set -eux; \
...
+# Copilot SDK/CLI is excluded for the reason given further down (the copilot-native
+# provider is unsupported in this image). An edge named here is never followed, at
+# any depth, and every store entry left unreached is deleted — so a new transitive
+# dependency of an excluded package cannot sneak back in behind a stale glob.
+COPY scripts/lib/prune-pnpm-store.mjs /tmp/prune-pnpm-store.mjs
+RUN node /tmp/prune-pnpm-store.mjs --store node_modules/.pnpm \
+      --importer /app --importer /app/server --importer /app/packages/shared \
+      --exclude prisma --exclude typescript \
+      --exclude @github/copilot --exclude @github/copilot-sdk \
+ && rm /tmp/prune-pnpm-store.mjs
```
mgcronin, line 173: **Advisory (A7), pre-existing, tracked in #39.** The legacy glob block just above (lines 136–152, unchanged) still runs *before* this prune and deletes `@prisma+debug@*`. That package is reachable (`@prisma/driver-adapter-utils` → `@prisma/debug`), so this reachability prune would have kept it. It is the cause of 428 of 917 `server/dist` modules failing to import, and the same happens on `main`. The PR's own argument, "prune by reachability rather than by glob … a glob that no longer matches fails silently", applies equally to a glob that matches too much. With this prune in place, most of that block (`prisma@*`, `@prisma+config`, `effect`, `c12`, `jiti`, …) is redundant, since those are reachable only via the excluded `prisma`. Removing the block would fix part of #39 and leave one source of truth. It's reasonable to do that in #39 rather than here.

Dockerfile.server
```diff
@@ -156,6 +157,48 @@ RUN set -eux; \
...
+#     linuxmusl-<this arch> (this is alpine/musl; some packages load the glibc one
+#     through gcompat).
+ARG TARGETARCH
+RUN set -eux; \
+    for D in $(find node_modules/.pnpm -maxdepth 3 -type d -path '*/node_modules/tree-sitter-*' 2>/dev/null); do \
+      rm -rf "$D/prebuilds" "$D/bindings" "$D/binding.gyp" 2>/dev/null || true; \
+      find "$D" -maxdepth 2 -type d -name src -prune -exec rm -rf {} + 2>/dev/null || true; \
+    done; \
+    for D in $(find node_modules/.pnpm -maxdepth 3 -type d -path '*/node_modules/better-sqlite3' 2>/dev/null); do \
+      rm -rf "$D/deps" "$D/src" "$D/build/Release/obj.target" "$D/build/Release/.deps" \
```
mgcronin, line 193: Advisory (A8), low. In the built image, `better-sqlite3@12.11.1/build/Release/` still holds `obj/` (10 MB) and `sqlite3.a` (3 MB). The prune targets `obj.target`, but this node-gyp output also writes `obj`. Adding `"$D/build/Release/obj" "$D/build/Release/sqlite3.a"` recovers about 13 MB.

.changes/unreleased/34-slimmer-server-image.md
```diff
@@ -0,0 +1,10 @@
+---
+issue: 34
+section: Fixed
+---
+
+- The `metis-server` image no longer ships the Copilot CLI binary, the Prisma CLI's
+  dependency tree, tree-sitter C sources and native builds, compile-time SQLite
+  sources, other platforms' prebuilt binaries, Oracle's JDBC jars, or npm/yarn —
+  roughly 1.3 GB down to about 0.9 GB on amd64. Its size gate now holds it to its
```
mgcronin: nit: CI measured 821 MB (the docs say ~820), so "about 0.8 GB" matches the measurement better than "about 0.9 GB", which reads like the budget.

scripts/lib/verify-image-size.mjs
```diff
@@ -23,7 +24,20 @@ import process from "node:process";
 
 export const DEFAULT_MAX_IMAGE_MB = 350;
 
-/** @typedef {{ tag: string, mb: number | null, exempt: boolean }} ImageSize */
+/**
+ * `metis-server`'s own budget, in MB, measured on a GitHub-hosted amd64 runner (#34).
+ *
+ * The general 350 MB is unreachable for the server without dropping features: the
```
mgcronin: **Non-blocking (residual B2), low severity, high confidence.** This JSDoc still states the retired rationale as fact: "unreachable for the server without dropping features", counting the Oracle Instant Client and LanceDB. #39 shows neither loads in this image. It then sends the reader to docs/OPERATIONS.md for "the reasoning", and that section now says the opposite: a measured ceiling, provisional until #39. Suggest rewording to match, e.g. "Measured 821 MB + ~10% headroom; a regression ceiling, provisional until #39 (Instant Client and LanceDB do not load today). See docs/OPERATIONS.md."

.github/workflows/ci.yml
```diff
@@ -118,11 +111,32 @@ jobs:
...
+
       # Issue #145 — image-size budget gate. The `metis-embeddings` sidecar is
       # built but exempt from the cap (see docs/OPERATIONS.md > "Container
       # Image Sizes").
       #
-      # Budget note: the documented amd64 floor is ~329 MB, gated at ≤350 MB.
+      # Budget note (#34): metis-ui is gated at ≤350 MB. metis-server is held to
+      # its OWN measured amd64 budget (900 MB, DEFAULT_MAX_SERVER_IMAGE_MB) —
+      # the old ~329 MB figure was never re-measured as the image grew, and
+      # 350 is unreachable for it without dropping features. The breakdown is in
```
mgcronin: **Non-blocking (residual B2), low, high.** This has the same stale claim as the `DEFAULT_MAX_SERVER_IMAGE_MB` JSDoc: "350 is unreachable for it without dropping features". It contradicts the provisional wording now in docs/OPERATIONS.md. Please reword it the same way.

docs/OPERATIONS.md
```diff
@@ -405,10 +452,46 @@ pnpm verify:image-size
...
+# Tighten the budgets (e.g. after the LanceDB sidecar lands)
+MAX_IMAGE_MB=250 MAX_SERVER_IMAGE_MB=700 pnpm verify:image-size
 ```
 
+### CI build cache (#3)
+
+`ci.yml`'s `api` job builds the four images with `docker/build-push-action`
+and caches layers in the GitHub Actions cache (`type=gha`, one `scope` per
+image, `mode=max` so the multi-stage builders' `pnpm install` layers are cached
+too). `build-images.yml` uses the same scopes, but its release-tag build
```
mgcronin, line 464: **Non-blocking, low, high.** "`build-images.yml` uses the same scopes" is accurate only for `server` and `ui`. The build-images matrix names are `embeddings-svc` and `copilot-svc`, which gives scopes `metis-embeddings-svc` and `metis-copilot-svc`. `ci.yml` writes `metis-embeddings` and `metis-sql-lineage`. Now that the tag build is read-only (correctly), and since the PR-validation step never runs (see the build-images.yml comment), nothing ever writes the `-svc` scopes. Those two release builds are therefore always cold. This is not a regression: the cache they used to write could only be read by the same tag. Either say "server and ui share scopes with ci.yml", or align the scope names (e.g. `scope=metis-embeddings` for the embeddings entry).

.github/workflows/build-images.yml
```diff
@@ -163,8 +174,14 @@ jobs:
           # PR builds don't push to GHCR so provenance is meaningless here.
           provenance: false
           sbom: false
-          cache-from: type=local,src=/tmp/buildkit-cache/${{ matrix.image.name }}
-          cache-to: type=local,dest=/tmp/buildkit-cache/${{ matrix.image.name }},mode=min
+          # #3 — `type=gha` survives the runner; the old local-path cache in
+          # /tmp/buildkit-cache lived on the self-hosted box and is an empty
+          # directory on every hosted VM. Same scope as ci.yml's `api` build of
+          # this image, so the cache `main` writes warms this build's amd64
```
mgcronin, line 180: **Non-blocking, low, high; the gating predates this PR.** This step runs only `if: github.event_name == 'pull_request'`, but this workflow triggers only on `push: tags` and `workflow_dispatch`. It never runs. (Under `workflow_dispatch`, neither build step runs.) The new comment describes cache behaviour for a build that does not execute. The wiring test "the native validation build still writes the gha cache" (image-build-wiring.test.mjs:141) pins a `cache-to` on dead code, so it cannot catch a real cache regression. Suggest a follow-up issue to either delete the step or note that it is dormant. It does not block this PR.

mgcronin (commented):
Recommendation: merge with follow-ups. 0 blocking findings, 4 non-blocking.

This is an automated re-review (cycle 2), published as a COMMENT. It is not a human sign-off, and the reviewing identity is also the PR author. It covers only what commit 487f261 changed. I did not re-review the whole PR.

### B1: resolved
All three rows trace to the cited jobs. I pulled step timings with `gh api repos/openzigs/metis/actions/jobs/<id>` for 106567592133, 106574086399 and 106589624773. Every figure matches to the second:
- server, ui, embeddings, sql-lineage
- four builds: 13m30s / 7m45s / 7m25s
- whole job: 28m20s / 23m22s / 18m46s
- Test: 8–11 min. Free disk: 45s–1m49s

The job logs also confirm the 83 GB → 110 GB `df` figures and `metis-server … 821 / 900` from the size gate. The PR body is no longer a stub. Issue #3's before/after warm measurement is present. One caveat: "warm" here means a re-run on the PR ref. A tag or PR seeded from `main`'s cache was not measured.

### B2: resolved in the doc, but the same claim is still in two code comments (non-blocking)
`docs/OPERATIONS.md` now says the budget is a measured ceiling, provisional until #39. It names the DPI-1047 failure and the LanceDB/musl failure (~237 MB) and asks for a re-measure. The two table rows are annotated. That is accurate. However, the retired rationale ("350 is unreachable … without dropping features") is still stated as fact in the JSDoc on `DEFAULT_MAX_SERVER_IMAGE_MB` and in the `ci.yml` gate comment. Both point readers to the doc, which now says the opposite. See the inline comments. My cycle-1 B2 comment named only the doc, so I am not escalating this, but it is a two-line fix worth making before merge.

### A4: correct, and the new test can fail
The tag build keeps `cache-from: type=gha,scope=metis-${{ matrix.image.name }}`. A tag run can restore default-branch caches, and `ci.yml` (push to `main`) writes `metis-server` and `metis-ui`, so the release build can still read the cache for those two. I checked falsifiability with mutations on a scratch copy; the worktree was not touched:
- Re-adding `cache-to` to the tag step makes 2 tests fail.
- Dropping the tag step's `cache-from` makes 2 tests fail.
- Changing `mode=max` to `mode=min` on the PR step makes 1 test fail.
- Restoring the file brings all 18 back to passing.

### Agent memory: clean
The two new files and `MEMORY.md` contain no absolute paths, hostnames, IPs, emails or private-terms matches. I also checked the whole PR diff against the private-terms list and found nothing. `pnpm agents:verify` exits 0 (store: 2 files, 485 B index; the longest entry is ~143 bytes against the 150 cap). `pnpm changelog:verify` exits 0.

### CI
All checks pass on head 487f261: api, api-outcome, postgres-adapter, ui, CodeQL, Semgrep and the rest. e2e and windows were skipped.

### Non-blocking findings (inline)
1. The budget rationale is still stale in the `verify-image-size.mjs` JSDoc and the `ci.yml` comment (low severity, high confidence).
2. "`build-images.yml` uses the same scopes" is true only for server and ui. `embeddings-svc` and `copilot-svc` use scopes that nothing writes, so their release builds are always cold (low, high).
3. The PR-validation step in `build-images.yml` is gated on `pull_request`, which is not a trigger of that workflow, so it never runs. This predates the PR. The new comment and the test "native validation build still writes the gha cache" describe and pin a step that never executes (low, high).
4. Informational: embeddings did not benefit from the warm cache (2m11s cold, 2m12s warm). The aggregate "roughly halves" is still true.

Issue #25, opened 2026-09-21
Documentation generation on a thinking-by-default model: 72 minutes and truncated sections


mgcronin:
**A full-project Architecture document for this repository took 72 minutes on `deepseek-v4-pro`, and came back degraded: three of its section groups were truncated by the 8,192-token output cap.** The retrieval core — LanceDB, BM25 hybrid search, the code graph — is never mentioned in the result.

_Found during an end-to-end walkthrough of a fresh local install (SQLite, Node 22.22, `AI_PROVIDER=anthropic` pointed at DeepSeek's Anthropic-compatible endpoint, `ANTHROPIC_MODEL=deepseek-v4-pro`), generating documentation for the `openzigs/metis` repository itself._

## Measured

| | |
|---|---|
| Scope / type | Full Project / Architecture, web research off |
| Modules (phase 1) | 174 — about 23 s each, ~62 min |
| Phase 2 synthesis | ~10 min |
| Model calls | 178 |
| Tokens | 1,777,445 input · 1,375,813 output · 131,584 cache-read |
| Result | 27,049 chars / 3,368 words, status `degraded` |
| Truncated groups | Components & Data Architecture · Cross-Cutting Concerns & Integrations · Algorithms, Operations & Stack |

What the document got right: Express, Prisma, Next.js, Socket.IO, SQLite/Postgres, the sqlglot sidecar, and the Copilot SDK — no invented frameworks. The degraded state and the three truncation warnings are shown clearly in the UI, which is the right behaviour.

## Likely cause of truncation

Output is several times larger than the text it produces — a phase-1 fact extraction for a two-line migration module returned 1,275 output tokens for 116 input. DeepSeek runs in thinking mode by default, and reasoning tokens appear to be drawn from the same `max_tokens` budget as the section text, so an 8,192 cap leaves too little for the answer. (The Anthropic provider already accounts for thinking inside `max_tokens` for Claude, #1257; the budget itself is not adjusted for a model that always thinks.)

## Also seen

- `Faithfulness judge batch was unparseable; treating batch as unverifiable` — one judge batch lost.
- `Section facts exceeded the facts char cap — modules omitted` for the ops-and-stack section.

## Acceptance criteria

- [ ] The per-section output budget accounts for reasoning tokens (or thinking is disabled / reduced for docs-gen on providers that support it), so sections are not truncated on a thinking-by-default model
- [ ] Phase 1 runs modules concurrently up to a configurable limit, so a 174-module project does not take an hour
[...]

PR #41, opened 2026-09-22
fix: Resolve #22, #25 — unpriced models recorded as unpriced; docs-gen fits thinking-by-default models

### LINKED ISSUE #22: Cost tracking records $0.00 for models with no configured price
**Usage from a model METIS has no price for is recorded with `estimatedCostUsd = 0`.** With `deepseek-v4-pro`, every call shows as free in the cost views, which reads as "no spend" rather than "unknown spend".

## Acceptance criteria

- [ ] Usage for an unpriced model is marked "unpriced" (null cost), never `0`
- [ ] Cost views show unpriced spend separately, with its token counts
- [ ] An administrator can configure per-model input/output/cache-read prices
- [ ] A test: recording usage for an unknown model yields a null cost, not 0

mgcronin:
Closes #22
Refs #25: this leaves #25 open. Its last criterion (the Architecture document covers the RAG subsystem) and the Phase-1 speed-up both need a live DeepSeek run, which has not been done. See "Not verified".

Both issues come from running METIS on a non-Claude provider (DeepSeek through `ANTHROPIC_BASE_URL`), so they ship as one PR.

## #22: one pricing source, and unpriced is not $0

**Before:** there were two rate tables that disagreed. `token_usages` (docs-gen, finops `recordUsage`) priced any unknown model on the `anthropic` provider at **Sonnet 4.6** rates through an `anthropic:default` row. `ai_token_usages` (chat, impact analysis, `TokenTracker`) recorded the same model at **0**. The two tables also disagreed on Opus 4.8: $5/$25 in one, $15/$75 in the other.

**Now:**
- `resolveRate(provider, model)` in `server/src/lib/finops/provider-rates.ts` is the only price lookup, and both trackers use it. It checks, in order:
  1. the administrator's `MODEL_PRICES` setting;
  2. built-in `provider:model` rows;
  3. a Claude family match on the model id: Opus 4.5+ at $5/$25, Opus 4/4.1 at $15/$75, Sonnet 5 at $2/$10, other Sonnets at $3/$15, Haiku at $1/$5. These prices were checked against Anthropic's pricing page on 2026-09-21;
  4. a `provider:default` row, which exists only for the free internal providers `offline-stub` and `copilot-native`.
  
  If none of these matches, the result is `null` (**unpriced**). `MODEL_PRICES` entries are keyed by `model` or `provider:model`, in USD per MTok.
- **Unpriced usage is stored as NULL.** `token_usages.costCents` is now nullable. There are migrations for SQLite (table rebuild, generated by `prisma migrate diff`) and Postgres (idempotent `DROP NOT NULL`/`DROP DEFAULT`), and the cumulative Postgres init baseline is updated to match. `ai_token_usages.estimatedCostUsd` was already nullable.
- **Anthropic list prices are not applied when `ANTHROPIC_BASE_URL` points at a host other than `api.anthropic.com`.** [DeepSeek maps `claude-haiku-*` to `deepseek-flash` and `claude-opus-*` to `deepseek-v4-pro`](https://api-docs.deepseek.com/guides/anthropic_api), so a `claude-*` name on that endpoint is not a Claude price. Admin prices still apply.
- **Administrators can set prices** with the new `MODEL_PRICES` tunable (JSON, validated with strict zod). It is written through the existing admin config surface (`PUT /api/admin/config/:key`, `admin.write`) and can be edited in Settings → Configuration.
- **The cost views show unpriced usage separately.**
  - `summarizeUsage()` and `UsageService` return the priced cost plus `unpriced` token totals, and each row carries `unpricedTokens`. A group with no priced usage has a `null` cost.
  - The project usage page has an "Unpriced usage" card with input/output tokens and call count. Cost cells read `Unpriced`, and a mixed row reads `$x + N unpriced tokens`.
  - The admin usage page has an "Unpriced Tokens" tile and column.
  - In CSV, an unpriced group has an empty cost cell and a new `unpricedTokens` column.
- Priced-only aggregators (forecast, chargeback via `_sum`, AWS reconciliation, runs cost, eval runner, test-coverage estimate) treat NULL as "no known cost". Each one has a comment saying so.

## #25: docs-gen on a thinking-by-default model

- **Output budget accounts for reasoning.**
  - `deepseek-v4-pro` was not in the docs-gen ceiling table, so the section cap fell to the 8,192 unknown-model default. That is the value #25 truncated at.
  - I added DeepSeek V4's documented ceiling: `max_tokens` "must be between 1 and 384K (393216)", per the [chat-completion API reference](https://api-docs.deepseek.com/api/create-chat-completion) and the Models & Pricing page.
  - DeepSeek's thinking mode is ["enabled by default"](https://api-docs.deepseek.com/guides/thinking_mode), and its own default `max_tokens` is 8K without thinking but 64K with it. So models on a documented think-by-default list get `DOCS_GEN_REASONING_ALLOWANCE_TOKENS` (default 32,768) on top of every docs-gen answer cap. The sum is still clamped to the model ceiling.
  - A DeepSeek section now asks for 65,536 tokens (was 8,192). Other models are unchanged.
  - I did not disable thinking. Correction from review: DeepSeek's [thinking-mode guide](https://api-docs.deepseek.com/guides/thinking_mode) gives `{"thinking":{"type":"disabled"}}` for the **OpenAI** format. The Anthropic format uses `{"reasoning":{"effort":"none"}}`. Turning thinking off changes output quality and would need a live comparison.
- **Unparseable judge batch is retried once.** `FaithfulnessJudge.judgeBatch` retries only an *unparseable* response, and not when the caller has aborted. A batch that parses but matches too few claims is a different failure and is not retried.
- **Phase 1 runs as a concurrent pool.** Phase 1 was already concurrent, but in fixed `Promise.allSettled` batches. Each batch waited for its slowest module, and the limit could only be set as a raw `process.env` value.
  - It is now a worker pool (`mapSettledWithConcurrency`), so the next module starts as soon as any module finishes.
  - `DOCS_GEN_PHASE1_CONCURRENCY` is now a registry tunable (1–64, db → env).
  - The default stays **3**, as before. `docs/ARCHITECTURE.md` said 6, which was wrong, and is corrected.

## Tests: each one was shown to fail by mutating the implementation

| # | Mutation | Went red |
|---|---|---|
| M1 | put the `anthropic → Sonnet 4.6` fallback back (the pre-#22 behaviour) | 6 tests, incl. "records the #22 DeepSeek run as UNPRICED in BOTH tables" and the usage-summary and tick tests |
| M2 | unknown model → zero rate (`DEFAULT_RATE`) | 11 |
| M2b | drop the free-provider `provider:default` row | 1 |
| M3 | ignore the `MODEL_PRICES` override | 6 |
| M4 | drop the third-party `ANTHROPIC_BASE_URL` check | 3 |
| M5 | `estimateCostUsd` returns 0 for an unpriced model (pre-#22 `ai_token_usages`) | 5 |
| M6 | price legacy Opus 4/4.1 at the Opus 4.5+ rate | 2 |
| M7 | `summarizeUsage` treats a NULL cost as 0 | 1 |
| M8 | `UsageService` treats a NULL cost as 0 | 2 |
| M9/9b/9c | UI shows `$0.00` for unpriced; hide the unpriced card; admin page shows `$0.0000` | 1 each |
| M10 | no reasoning allowance | 6 (unit + synthesizer e2e) |
| M11 | no DeepSeek ceiling row | 6 |
| M12 | `JUDGE_BATCH_ATTEMPTS = 1` | 3 |
| M12b | retry even after an abort | 1 |
| M13 | pool with a single worker | 4 |
| M14 | Phase-1 ignores the configured limit | 2 (e2e) |
| M15 | pool replaced by the old fixed batches | 2, incl. the e2e test that holds module 0 open |
| M16 | `MODEL_PRICES` registry key without validation | 1 |
| M17 | override unit conversion ÷1000 instead of ÷10 | 1 |
| M18 | runs-service attributes cost to an unpriced model | 1 |
| M19 | override lookup without `Object.hasOwn` | 1 |

M19 found a real defect while I was writing it. A model id such as `constructor` resolved to an `Object.prototype` member and would have priced at NaN. The lookup is now own-keys only.

Several tests fail on the unfixed code by construction. `usage-consistency` and `finops-token-tracker` use the measured #22 figures (1,334,017 in / 1,297,372 out → 2,349¢ before, NULL now). The synthesizer e2e asserts a 65,536-token section cap (8,192 before).

## Coverage (vitest v8, touched files, same test set before and after)

|---|---|---|
| finops/provider-rates.ts | 100% → 98.0% | 100% → 96.7% |
| finops/token-tracker.ts | 93.9% → 93.9% | 100% → 100% |
| usage/usage-service.ts | 97.9% → 98.2% | 85.7% → 92.0% |
| docs-gen/output-caps.ts | 94.4% → 100% | 93.8% → 100% |
| docs-gen/grounding/faithfulness-judge.ts | 100% → 100% | 90.5% → 91.3% |
| ai/token-tracker.ts | 80.6% → 79.7% | 72.7% → 72.7% |
| docs-gen/phase1-concurrency.ts (new) | — → 100% | — → 100% |
| finops/model-prices-schema.ts (new) | — → 100% | — → 100% |
| replay/runs-service.ts | → 100% lines (`tests/replay-runs-service.test.ts`) | |

`ai/token-tracker.ts` is flat because its uncovered lines are `dailyRollup` and `hashPrompt`, which this PR does not touch. The changed functions (`estimateCostUsd`, `estimateUsageCostUsd`) are fully covered.

## Gate (run locally before pushing)
- `pnpm lint`: exit 0. This includes the private-terms, licence and CLA scans, run after staging.
- `pnpm typecheck`: exit 0.
- Tests, run per package because the root `pnpm test` fan-out exceeds my 10-minute command limit:
  - server: 1104 files passed / 3 skipped, 17,207 tests;
  - ui: 351 files, 3,242 tests;
  - `@metis/shared`: 36 files, 698 tests.
- `node scripts/verify-changelog-fragment.mjs`: passes. There are two fragments: `22-unpriced-model-cost.md` and `25-docs-gen-thinking-models.md`.
- `prisma migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma`: empty.

## Security (OWASP pass on the diff)
- `MODEL_PRICES` is admin-only (`admin.write`). It is validated with a strict zod schema: non-negative bounded numbers, unknown fields rejected, invalid JSON rejected. It is re-validated on read and ignored rather than thrown on if invalid. Lookups are own-key only (M19).
- No SQL built from input and no shell use. `ANTHROPIC_BASE_URL` is only parsed with `new URL` to read the host.
- No new dependencies, so no dependency audit delta.

## Found along the way, not fixed here
- `bedrock-gateway:us.anthropic.claude-fable-5` is priced at the Haiku tier, and `us.anthropic.claude-sonnet-5` at the 4.x Sonnet rate. Bedrock pricing is partner-set and I did not verify it, so the Fable row is filed as #42. The direct-API Fable 5 / 5.1 prices were added in the review round.
- On DeepSeek, the anthropic docs-gen claim-extraction model defaults to `claude-haiku-4-5`, which DeepSeek maps to `deepseek-flash`. Its usage is recorded under the Claude name, and with this PR it is correctly unpriced.
- The project usage page's header comment says it listens for `usage:tick`. No UI code subscribes, so the socket event has no consumer.
- Existing `token_usages` rows that were priced by the old Sonnet fallback are not rewritten. Which rows used the fallback cannot be recovered from the table.

## Not verified
- #25's last criterion, "the generated Architecture document covers the RAG/retrieval subsystem", needs a live docs-gen run against DeepSeek. I was told not to touch the dev server, so this is **unverified**.
- The Phase-1 wall-clock improvement is not measured either. The pool removes head-of-line blocking, but I could not reproduce the 62-minute run. Its per-module figure suggests the old loop was close to serial, and the cause may lie outside the batching.
- CI's `api` job has a known image-size failure (#34, fixed by #38). That is not this PR's to fix.

## Review round 1 (review 5273813725), commit 2b00b91

**Blocking, fixed. The autopilot cost ceiling now fails closed on unpriced usage.**
- `projectMonthlyCostForCeiling()` in `budget-enforcer.ts` re-prices NULL-cost rows with the current price source, so an administrator who sets `MODEL_PRICES` also covers the month's earlier rows. It returns the tokens that stay unpriced.
- `assertAutopilotAllowed` and `assertCeiling` (the watchdog's check) throw `AutopilotCostCeilingError` with `unpricedTokens` when a ceiling is set and any month-to-date usage is unpriced. The message names `MODEL_PRICES`. With no ceiling set, behaviour is unchanged.
- The error is the existing class, with the same 402 status and `AUTOPILOT_COST_CEILING` code, so the runner's abort path and HTTP mapping need no change.
- Tests went red under both mutations:
  - no unpriced check: 2 tests;
  - no re-pricing: the MODEL_PRICES test, plus 2 older ceiling tests. Those two failed only because the mutated line calls `resolveRate` on mock rows that have no model.
- I traced the other cents-based limits. The PR-reviewer budget (`budget-guard.ts`) sums `costUsd` from the agent envelope, which this PR does not change. The test-coverage per-run budget is covered in #43.

**Advisory**
| # | Finding | Outcome |
|---|---|---|
| 2 | Proxy/gateway `ANTHROPIC_BASE_URL` becomes unpriced | Fixed. New tunable `ANTHROPIC_BASE_URL_BILLS_AS` (`auto` \| `anthropic`). The changelog fragment and USER_GUIDE flag the behaviour change. |
| 3 | Headline tiles show $0.00 | Fixed. Tiles read `Unpriced` or `$x + N unpriced tokens`. The summary gains `monthToDateUnpricedTokens` for the projection tile, and the admin Estimated Cost tile reads `Unpriced`. |
| 4 | local-gemma now unpriced | Fixed. `local-gemma:default` is back as a real $0. |
| 5 | Fable 5 unpriced on the direct API | Fixed. Fable 5 is $10/$50 with $1 cache hits; Fable 5.1 has $0.25 cache hits (pricing page, 2026-09-21). The Bedrock row is #42. |
| 6 | Reasoning allowance keyed on the requested name | Fixed. A `claude-*` name sent to DeepSeek's host now gets the allowance. The setting text states the non-streaming SDK clamp, and `v4-flash` is dropped. |
| 7 | Coverage tracker skips the provider check | Declined here, filed as #43. Passing the real provider would make the estimate `null` → `?? 0` on DeepSeek, which would switch off the per-run budget (the same shape as the blocking finding). The tracker also gets no provider where it is built. |
| 8 | Rollback comment | Fixed. Both migrations now say the rollback is unexecuted and lossy. |
| 9 | `Closes #25` | Fixed: now `Refs #25`. The default concurrency stays 3; see the thread. |
| 10 | Thinking-disable claim | Corrected in this body. |
| 11 | `MODEL_PRICES` parsed per record | Fixed. It is parsed once per distinct raw value. |

Every new test was red before its fix and under a targeted mutation after it.


server/src/lib/finops/provider-rates.ts
```diff
@@ -123,61 +161,187 @@ const RATES: ReadonlyMap<string, TokenRate> = new Map([
...
+export function resolveRate(
+  provider: string | undefined,
+  model: string,
+  opts: ResolveRateOptions = {},
+): TokenRate | null {
+  const config = opts.config ?? getConfigService();
+  const override = overrideRate(provider, model, config);
+  if (override) return override;
+
+  if (provider === "anthropic" && isThirdPartyAnthropicEndpoint(opts.env)) return null;
```
mgcronin: This check comes after the admin override and before every built-in row. DeepSeek's Anthropic endpoint maps `claude-haiku-*` to `deepseek-flash` and `claude-opus-*` to `deepseek-v4-pro` (https://api-docs.deepseek.com/guides/anthropic_api). So on a third-party base URL, a `claude-*` model name says nothing about the price, and pricing it at Anthropic's list rate is the same fault as the old Sonnet fallback. An unparseable URL also counts as third-party, because we cannot tell whose prices apply.

server/prisma/migrations/20260921000000_issue22_token_usage_unpriced_cost/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+    "projectId" TEXT NOT NULL,
+    "sessionId" TEXT,
+    "provider" TEXT NOT NULL,
+    "model" TEXT NOT NULL,
+    "inputTokens" INTEGER NOT NULL DEFAULT 0,
+    "outputTokens" INTEGER NOT NULL DEFAULT 0,
+    "cacheReadTokens" INTEGER NOT NULL DEFAULT 0,
+    "cacheWriteTokens" INTEGER NOT NULL DEFAULT 0,
+    "totalTokens" INTEGER NOT NULL DEFAULT 0,
+    "costCents" INTEGER,
```
mgcronin, line 36: This is the SQLite table rebuild, taken verbatim from `prisma migrate diff`. `INSERT..SELECT` copies every existing `costCents` unchanged. Rows the old fallback priced at Sonnet rates are deliberately not rewritten, because the table does not record which rows used the fallback. The rollback is written in the header.

server/src/lib/docs-gen/output-caps.ts
```diff
@@ -119,7 +167,11 @@ export function resolveDocsGenMaxOutputTokens(
   const safeFallback = ceiling === null ? Math.min(fallback, UNKNOWN_MODEL_SAFE_DEFAULT) : fallback;
   const raw = config.getNumber(key, safeFallback);
   const configured = Number.isFinite(raw) && raw >= MIN_MAX_OUTPUT_TOKENS ? raw : safeFallback;
-  return ceiling === null ? configured : Math.min(configured, ceiling);
+  // #25 — the configured value is the ANSWER budget; a model that reasons by
+  // default spends its reasoning from the same `max_tokens`, so add room for it
+  // before clamping to the model's ceiling.
+  const withReasoning = configured + reasoningAllowanceTokens(model, config);
```
mgcronin, line 207: The configured value is now the answer budget, and reasoning headroom is added on top, only for models on the documented think-by-default list. The sum is clamped to the model ceiling afterwards, so this can never exceed what the model accepts. On the anthropic provider, non-streaming calls are additionally held to the SDK's 21,333 bound by `boundNonStreamingOutputTokens` (#1257).

server/src/lib/docs-gen/grounding/faithfulness-judge.ts
```diff
@@ -272,21 +278,37 @@ export class FaithfulnessJudge {
...
+        ...(this.responseFormat ? { responseFormat: this.responseFormat } : {}),
+      });
+
+    // #25 — an unparseable batch is retried ONCE before it is counted as
+    // unverifiable: a sampled model can emit a malformed or cut-off verdict list
+    // on one call and a clean one on the next, and each lost batch removes its
+    // claims from the section's faithfulness score. A batch that parses but
+    // matches too few claims is a different failure and is not retried.
+    let rawVerdicts: ClaimVerdict[] | null = null;
+    for (let attempt = 1; attempt <= JUDGE_BATCH_ATTEMPTS; attempt++) {
```
mgcronin, line 303: Only a parse failure is retried. A response that parses but matches under 50% of the claims is a model malfunction, not a transport glitch, and retrying it would double the judge cost for the SAS-style failure mode without evidence it helps. No retry happens after an abort.

server/src/lib/docs-gen/holistic-synthesizer.ts
```diff
@@ -1152,45 +1153,46 @@ export async function synthesizeHolisticDocument(
...
-  // Default concurrency is 3 to stay within ALB idle timeout limits when
-  // using Sonnet on large modules. Bump via DOCS_GEN_PHASE1_CONCURRENCY
-  // if your gateway supports longer timeouts.
+  // Phase 1: extract compact facts per module, keeping up to
+  // DOCS_GEN_PHASE1_CONCURRENCY extractions in flight (#25 — a worker pool, not
+  // fixed batches that each waited for their slowest module). Registry-backed
+  // (db → env), default 3; raise it for a provider/gateway that allows more.
   const facts: ModuleFacts[] = [];
-  const CONCURRENCY = Number(process.env.DOCS_GEN_PHASE1_CONCURRENCY) || 3;
+  const concurrency = resolvePhase1Concurrency();
```
mgcronin, line 1161: The default stays 3, the value the batch loop used. Raising it is an operator decision per provider (DeepSeek documents a 500-request concurrency limit for v4-pro; a gateway behind an ALB may not tolerate that). The pool is what removes the slowest-in-batch stall. The e2e test holds module 0 open and asserts every other module still ran.

mgcronin (commented):
Author notes on the load-bearing decisions: why each line is shaped the way it is. These are not an approval.

server/src/lib/finops/budget-enforcer.ts
```diff
@@ -73,7 +73,8 @@ async function getMtdAggregate(
   let cents = 0;
   for (const r of rows) {
     tokens += r.totalTokens;
-    cents += r.costCents;
+    // #22 — an unpriced row (null) adds tokens but no known cost.
+    cents += r.costCents ?? 0;
```
mgcronin: **BLOCKING, Medium (A04, a guard that cannot fire).** `getMtdAggregate` now counts only priced cents, and `projectMonthlyCost()` is what `assertAutopilotAllowed` and `assertCeiling` (`autopilot/autopilot-runner.ts:86,105`) compare against `autopilotCostCeilingCents`. This PR targets deployments (DeepSeek through `ANTHROPIC_BASE_URL`) where **every** row is NULL. On those, the projection is permanently 0 and the autopilot cost ceiling can never fire.

Before this PR the same deployment's docs-gen rows were priced at Sonnet rates, so the ceiling tripped early rather than never. Fixing the over-pricing (correct) has quietly switched a spend control off (not correct), and no test covers it.

Suggested fix, pick one:
- (a) If a ceiling is set and month-to-date usage includes unpriced rows, fail closed with a distinct error along the lines of "cost ceiling cannot be evaluated: N unpriced tokens; set MODEL_PRICES".
- (b) Return `unpricedTokens` from `getMtdAggregate` and have the ceiling check refuse when it is non-zero.

Either way, add a test in which an all-unpriced project with a ceiling does **not** pass `assertAutopilotAllowed`.

server/src/lib/finops/provider-rates.ts
```diff
@@ -123,61 +161,187 @@ const RATES: ReadonlyMap<string, TokenRate> = new Map([
...
-  if (/^claude-haiku-4/.test(model)) return ANTHROPIC_HAIKU_4;
-  return undefined;
+export function isThirdPartyAnthropicEndpoint(
+  env: NodeJS.ProcessEnv | Record<string, string | undefined> = process.env,
+): boolean {
+  const raw = env.ANTHROPIC_BASE_URL?.trim();
+  if (!raw) return false;
+  try {
+    const host = new URL(raw).hostname.toLowerCase();
+    return host !== "api.anthropic.com";
```
mgcronin, line 246: Medium, advisory. Any host other than `api.anthropic.com` counts as third-party. That includes the common case of `ANTHROPIC_BASE_URL` pointing at a corporate egress proxy, LiteLLM, or an AI gateway that relays to Anthropic. After this PR every Claude call on such a deployment is recorded as unpriced, where before it was correctly priced. `MODEL_PRICES` can restore it, but only by re-entering every Claude price by hand. The key-registry text says so, but the changelog fragment does not warn upgraders.

Consider an explicit override, e.g. `ANTHROPIC_BASE_URL_BILLS_AS=anthropic|other`, or keying the skip on the documented DeepSeek host. At minimum, call it out in `22-unpriced-model-cost.md` as a behaviour change.

server/src/lib/finops/provider-rates.ts
```diff
@@ -123,61 +161,187 @@ const RATES: ReadonlyMap<string, TokenRate> = new Map([
...
+  if (provider) {
+    const exact = RATES.get(`${provider}:${model}`);
+    if (exact) return exact;
+  }
+  const family = claudeFamilyRate(model);
+  if (family) return family;
+  // A provider-wide row exists only for METIS-internal providers whose calls
+  // genuinely cost nothing per token (offline-stub, copilot-native). There is
+  // deliberately no such row for a paid provider (#22).
+  return (provider && RATES.get(`${provider}:default`)) || null;
```
mgcronin, line 304: Low, advisory. Only `offline-stub` and `copilot-native` keep a zero row. `local-gemma`, the self-hosted Ollama/vLLM provider (`ProviderKey` in `ai/types.ts`), used to fall through to `DEFAULT_RATE` ($0) and now resolves to `null`. Every local run will show as "Unpriced" and trigger the "set MODEL_PRICES" hint, though its per-token price is known to be zero. If that is intended, say so in the changelog; otherwise add `["local-gemma:default", DEFAULT_RATE]`.

server/src/lib/finops/provider-rates.ts
```diff
@@ -123,61 +161,187 @@ const RATES: ReadonlyMap<string, TokenRate> = new Map([
...
  *
- * Standard input/output tokens use the headline rate. Cache-read is usually
- * the cheapest tier (Bedrock advertises ~10% of input). Cache-write is more
- * expensive than vanilla input (Anthropic charges 1.25× input). Both
- * default to the input rate when the provider doesn't publish a separate
- * cache rate so we don't silently under-bill.
+ * Opus 4.5 and later are $5/$25; Opus 4 / 4.1 (and Claude 3 Opus) are
+ * $15/$75. Sonnet 5 is $2/$10; every earlier Sonnet is $3/$15.
+ */
+export function claudeFamilyRate(model: string): TokenRate | undefined {
```
mgcronin, line 325: Low, follow-up. `claudeFamilyRate` has no Fable branch, so `anthropic:claude-fable-5` on the direct provider is now unpriced, although Anthropic lists Fable 5 at $10/$50 (platform.claude.com pricing page, checked today). Unpriced is the honest failure mode, so this is not a correctness bug, but a first-party model with a known price should have a row. The Bedrock `claude-fable-5` → Haiku mis-row the PR body mentions is still there, and deserves an issue.

mgcronin (commented):
## Code Review Summary — PR #41

Recommendation: do not merge — 1 blocking finding

This is an automated analysis published as a comment. It is not a human sign-off. The reviewing identity (`mgcronin`) also authored the PR, so GitHub cannot record an approval or a change request from it. A human still decides whether to merge.

### Requirements: #22 PASS, #25 PARTIAL
- [x] #22: an unpriced model's cost is recorded as NULL in both `token_usages` and `ai_token_usages`, never 0 and never another model's price (`pricing-single-source.test.ts`)
- [x] #22: one pricing source, `resolveRate()`, feeds both tables
- [x] #22: the cost views show unpriced tokens separately. The headline tiles still read `$0.00` when all usage is unpriced (see inline comment)
- [x] #22: administrators can set per-model prices (`MODEL_PRICES`, validated, `admin.write`). The admin write is read back through `resolveRate` in `model-prices-config.test.ts`
- [x] #25: section output budget includes reasoning headroom. Section synthesis and Phase-1 facts use `stream()`, so the Anthropic SDK's 21,333 non-streaming bound does not clamp the 65,536 request
- [x] #25: an unparseable judge batch is retried once
- [~] #25: the Phase-1 pool is configurable, but the default is still 3 and the speedup was not measured
- [ ] #25: "the Architecture document covers the RAG subsystem" is unverified (it needs a live run). `Closes #25` will auto-close the issue with this criterion open. Use `Refs #25`, or split that criterion into its own issue.

### Security: 1 finding (A04, insecure design, a disabled cost control)
| Severity | Count | Details |
|---|---|---|
| Medium | 1 | The autopilot cost ceiling can no longer fire on unpriced usage (`budget-enforcer.ts:77`). **BLOCKING** |
| Low | 0 | — |

`MODEL_PRICES` is admin-only, validated with a strict zod schema and looked up by own keys only. It adds no SQL, no shell use and no dependencies. `ANTHROPIC_BASE_URL` is only parsed with `new URL` to read the host.

### Design / quality: GOOD, with advisories
One lookup, a clear resolution order, and a documented reason at each step. Advisories are inline:
- A proxy in front of Anthropic is treated as a third-party endpoint.
- `local-gemma` is now recorded as unpriced.
- The reasoning allowance is keyed on the model name METIS requests, not the model the endpoint serves.
- `testcoverage/cost-tracker` calls the price lookup without a provider.
- The migration's rollback is described in a comment but not implemented or tested.

### Performance: CLEAN
nit: `resolveRate` runs zod over `MODEL_PRICES` on every usage record. The cost is small; a cache keyed on the raw string would remove it.

### Tests: SUFFICIENT
I re-ran the tests and reverted production code myself rather than taking the author's table on trust:
- Reverting `output-caps.ts` to main: 10 tests failed.
- Reverting `faithfulness-judge.ts` to main: 3 failed.
- Reverting `holistic-synthesizer.ts` to main (the old fixed batches): the pool e2e test failed.
- Putting back the `anthropic:default` Sonnet row: 8 failed.
- Removing the third-party-endpoint check: 3 failed.
- Making `estimateCostUsd` return 0 for an unpriced model: 5 failed.
- Making the UI's `formatCents(null)` return `$0.00`: 1 failed.
- Touched suites green: server 17 files / 228 tests; the two UI usage-page files, 11 tests.

The author's mutation claims hold. No test covers the autopilot ceiling when usage is unpriced, which is how the blocking finding got through.

### Documentation: UP TO DATE
Two `.changes/unreleased/` fragments, no `CHANGELOG.md` edit. ARCHITECTURE, OPERATIONS and USER_GUIDE are updated, and the stale `DOCS_GEN_PHASE1_CONCURRENCY=6` row is fixed.

Informational: the PR body says DeepSeek's *Anthropic* format disables thinking with `{"thinking":{"type":"disabled"}}`. DeepSeek's thinking-mode guide lists that as the **OpenAI** format; the Anthropic format is `{"reasoning":{"effort":"none"}}`. This affects the PR body only, not the code. Correct it before anyone builds on it.

### Provider claims I checked against the provider's docs (2026-09-21)
- DeepSeek `max_tokens` "must be between 1 and 384K (393216)"; default 8K without thinking, 64K with it. Source: create-chat-completion reference.
- DeepSeek thinking is "enabled by default, with the default effort being `high`". Source: thinking-mode guide.
- DeepSeek maps `claude-opus*` → `deepseek-v4-pro`, and `claude-haiku*`/`claude-sonnet*` → `deepseek-flash`. Source: anthropic_api guide.
- Anthropic list prices: Opus 4.5–4.8 and Opus 5 $5/$25; Opus 4/4.1 $15/$75; Sonnet 5 $2/$10; Sonnet 4.x $3/$15; Haiku 4.5 $1/$5; Fable 5 $10/$50. Source: platform.claude.com pricing page.

### CI
Every check passes except `api`/`api-outcome`. `gh run view 35680318823 --json jobs` shows Lint, Typecheck and Test all succeeded; only "Verify image size budget" failed (`metis-server = 1301 MB exceeds 350 MB`). That is the pre-existing #34 (fix in open PR #38), not this PR. `e2e` and `windows` were skipped.

### Scope of this review
- Read both issues, including #22's correction comment.
- Read the full diff (46 files); read the pricing, budget, docs-gen cap, judge and provider files in full at the PR head.
- Local checks in a detached worktree at `4856d2c`: `pnpm typecheck` exit 0; eslint on the changed .ts/.tsx files exit 0; the targeted vitest suites and the mutations above.
- Not done: the full `pnpm test` fan-out (CI's `api` job ran it green), the running application, and a live DeepSeek docs-gen run.

server/src/lib/docs-gen/output-caps.ts
```diff
@@ -79,8 +79,56 @@ const MODEL_OUTPUT_CEILINGS: ReadonlyArray<readonly [RegExp, number]> = [
...
+ * enabled by default, with the default effort being high" for both V4 models
+ * (https://api-docs.deepseek.com/guides/thinking_mode), and its own default
+ * `max_tokens` is 8K in non-thinking mode but 64K in thinking mode — the
+ * reasoning is generated output. A cap sized for the ANSWER alone therefore
+ * truncates the answer (#25: three section groups cut at 8,192).
+ *
+ * Deliberately a short allow-list: a model is only given the extra budget on
+ * documented evidence that it thinks by default.
+ */
+const THINKING_BY_DEFAULT_MODELS: ReadonlyArray<RegExp> = [/deepseek-(v4-pro|v4-flash|flash)/i];
```
mgcronin: Low, advisory. The allowance is keyed on the model name METIS **requests**, not the model the endpoint **serves**. On DeepSeek's Anthropic endpoint, `claimModel` defaults to `claude-haiku-4-5` (`holistic-synthesizer.ts:421-425`), which DeepSeek serves as `deepseek-flash`, a thinking-by-default model. It matches the Claude ceiling row and gets no allowance.

Separately, claim and judge run through non-streaming `chat()`, which `AnthropicProvider.chat()` clamps to the SDK's 21,333 bound (`nonstreaming-output-bound.ts`). So the judge's nominal 65,536 is 21,333 on the wire. That is still more than 8,192, but the `DOCS_GEN_REASONING_ALLOWANCE_TOKENS` description ("claim/judge") overstates what is sent. Consider applying the allowance to any model on the `anthropic` provider when `isThirdPartyAnthropicEndpoint()` is true, and mention the non-streaming clamp in the description.

nit: `v4-flash` is not a model id DeepSeek documents; the documented ids are `deepseek-flash` and `deepseek-v4-pro`.

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -110,7 +110,9 @@ export class CoverageCostTracker {
         this.suggestionTokens += total;
         break;
     }
-    this.estimatedUsd += estimateCostUsd(modelId, prompt, completion);
+    // #22 — an unpriced model (null) has no estimable cost; its tokens are
+    // still counted above.
+    this.estimatedUsd += estimateCostUsd(modelId, prompt, completion) ?? 0;
```
mgcronin, line 115: Low, advisory. This call passes no `provider`, so `resolveRate(undefined, model)` skips the third-party-endpoint check and a `claude-*` model id is priced at Anthropic list rates even behind DeepSeek. That is the #22 over-pricing, surviving in this one estimate. The provider is available where the tracker is built, so pass it through.

ui/src/app/(authed)/projects/[id]/usage/page.tsx
```diff
@@ -132,6 +143,18 @@ export default function ProjectUsagePage() {
         />
       </section>
 
+      {u.unpriced.totalTokens > 0 ? (
```
Comment (mgcronin) [inline ui/src/app/(authed)/projects/[id]/usage/page.tsx:161]: Low-Medium, advisory. The separate card is good. The headline tiles just above it still render `formatCents(u.costCents)` and `formatCents(u.projectedMonthlyCostCents)` (line 138 onward), both numeric, so an all-unpriced window shows **Window cost $0.00** and **Projected month $0.00**. That is the "reads as no spend" symptom #22 opened with. The admin page's "Estimated Cost" tile does the same (`$0.0000`).

When `costCents === 0 && unpriced.totalTokens > 0`, render "Unpriced" or `$0.00 + N unpriced tokens`, as the provider-table cells already do.

server/prisma/migrations/20260921000000_issue22_token_usage_unpriced_cost/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+-- NULL now means "METIS had no price for this model when the row was written";
+-- the usage views show such rows separately with their token counts.
+--
+-- SQLite cannot relax a NOT NULL in place, so Prisma rebuilds the table (taken
+-- verbatim from `prisma migrate diff`). No existing row changes value: the
+-- INSERT..SELECT copies every costCents across unchanged. Rows priced by the
+-- old fallback are NOT rewritten — which rows were fallback-priced is not
+-- recoverable from the table.
+--
+-- Rollback: rebuild the table with `"costCents" INTEGER NOT NULL DEFAULT 0`,
```
mgcronin: Low. The rollback is written as a comment but neither implemented nor executed. It is also lossy: `COALESCE(costCents, 0)` turns unpriced rows back into $0, reintroducing the bug. The project seems to keep rollbacks as comments (1 of 115 SQLite migrations carries one), so this is not blocking. If anyone relies on it, it needs a test run against a copy of `dev.db`, and a note that it re-creates the $0 rows #22 set out to remove.

server/src/lib/finops/budget-enforcer.ts
mgcronin: Fixed in 2b00b91. The ceiling now fails closed.

- `projectMonthlyCostForCeiling()` (`budget-enforcer.ts`) re-prices each NULL-cost row with the current price source, so an administrator who sets `MODEL_PRICES` also covers the month's earlier rows. It returns the month-to-date tokens that are still unpriced.
- `assertAutopilotAllowed` and `assertCeiling` (and so the watchdog) go through one `checkCeiling()`. When a ceiling is set and `unpricedTokens > 0`, it throws `AutopilotCostCeilingError`. The error carries `unpricedTokens`, keeps the same 402 status and `AUTOPILOT_COST_CEILING` code, and its message says the ceiling cannot be evaluated and names `MODEL_PRICES`. With no ceiling set, behaviour is unchanged.
- Tests in `server/tests/autopilot-runner.test.ts` cover four cases: an all-unpriced project with a ceiling is refused; no ceiling still allows it; a mid-run `assertCeiling` throws; and with `MODEL_PRICES` set, the re-priced rows let the run pass under a high ceiling and trip a low one.
- Mutations: removing the unpriced check turns 2 tests red; removing the re-pricing turns the MODEL_PRICES test red.
- Following your durable finding, I checked the other cents-based limits. `pr-reviewer/budget-guard.ts` sums the agent envelope's `costUsd`, which this PR does not change. The test-coverage per-run budget has the same trap if its provider is fixed, so it is filed as #43.

server/src/lib/finops/provider-rates.ts
mgcronin: Fixed in 2b00b91 with an explicit override, as you suggested: a new tunable `ANTHROPIC_BASE_URL_BILLS_AS` (`auto` | `anthropic`, default `auto`).

- With `anthropic`, the built-in Claude prices apply behind a proxy or gateway.
- An unknown model is still unpriced under `anthropic`.

I kept `auto` as the host check rather than keying it on the DeepSeek host, because any other Anthropic-compatible provider has the same pricing problem. The `22-unpriced-model-cost.md` fragment and USER_GUIDE now describe this as a behaviour change and name the override.

Mutation: ignoring the override turns the new test red.

server/src/lib/finops/provider-rates.ts
mgcronin: Fixed in 2b00b91. `["local-gemma:default", DEFAULT_RATE]` is back. Self-hosted usage has no per-token charge, so it is a genuine $0, not unpriced. USER_GUIDE says so.

Mutation: dropping the row turns the new test red.

server/src/lib/finops/provider-rates.ts
mgcronin: Fixed in 2b00b91. I re-read the pricing page today and added two rows to `claudeFamilyRate`:

- Fable 5: $10 in / $50 out, $1 cache hit, $12.50 5-minute write;
- Fable 5.1: the same, except a $0.25 cache hit.

The Bedrock `claude-fable-5` → Haiku row is filed as #42. The page says Bedrock sets its own prices, so I did not copy the direct-API rate onto it.

Mutation: dropping the Fable branches turns the new test red.

server/src/lib/docs-gen/output-caps.ts
mgcronin: Fixed in 2b00b91. All three points:

- **Served model:** the reasoning allowance now also applies to a bare `claude-*` name when `ANTHROPIC_BASE_URL` is on `deepseek.com`, whose [Anthropic API page](https://api-docs.deepseek.com/guides/anthropic_api) maps opus names to `deepseek-v4-pro` and haiku/sonnet names to `deepseek-flash`.
  - I keyed it on the documented host rather than on any third-party endpoint, because a proxy relaying to Anthropic serves real Claude.
  - Bedrock `us.anthropic.claude-*` ids are excluded.
  - With that, the claim model `claude-haiku-4-5` on DeepSeek gets its facts/claim headroom.
- **Non-streaming clamp:** the `DOCS_GEN_REASONING_ALLOWANCE_TOKENS` description now says claim and judge calls are non-streaming, and on the anthropic provider are clamped to the SDK's 21,333 bound (or a lower per-model limit).
- **nit:** `v4-flash` is dropped from both regexes and from the test.

Mutations: removing the served-model check turns the DeepSeek test red; removing the `^claude-` guard turns the Bedrock-id test red.

server/src/lib/testcoverage/cost-tracker.ts
mgcronin: Declined in this PR and filed as #43, for two reasons.

1. **Fixing it here would disable the per-run coverage budget.** On DeepSeek, `claude-haiku-4-5` behind `anthropic` is unpriced. `estimateCostUsd` would then return `null`, and `record()`'s `?? 0` would stop the budget (`exceeded()` / `canAfford()`) from ever tripping. That is the same fail-open shape as the blocking ceiling finding. Today's over-estimate at Haiku list rates errs toward stopping early. The follow-up has to make unpriced usage fail closed in that budget, and that belongs with the provider fix.
2. **The provider is not available where the tracker is built.** `CoverageCostTracker` is constructed in `coverage-service.ts:101` from `runId`, `projectId` and `userId` only, and `record()` hard-codes `provider: "bedrock-gateway"` for the `ai_token_usages` row as well. Fixing this means threading the provider through from where the judge's `callModel` is wired.

I've left this thread open for you.

ui/src/app/(authed)/projects/[id]/usage/page.tsx
Comment (mgcronin) [reply ui/src/app/(authed)/projects/[id]/usage/page.tsx:161]: Fixed in 2b00b91.

- **Project page headline tiles:** they now read `Unpriced` when there is no priced usage, and `$x + N unpriced tokens` when usage is mixed, like the per-model cells.
- **Projected month tile:** the projection covers the month rather than the window, so `summarizeUsage` now returns `monthToDateUnpricedTokens` (server, shared schema, UI type) and the tile uses that.
- **Admin page:** the Estimated Cost tile reads `Unpriced` instead of `$0.0000` when all usage is unpriced. The Unpriced Tokens tile next to it already carries the count.

Mutations: reverting the tile formatter turns 2 tests red; removing the admin check turns 1 red; not counting MTD unpriced tokens turns the server summary test red.

server/prisma/migrations/20260921000000_issue22_token_usage_unpriced_cost/migration.sql
mgcronin: Fixed in 2b00b91. Both the SQLite and Postgres rollback comments now say they are documentation only, never executed or tested, and must be rehearsed on a copy of the database first. They also say the rollback is lossy: every unpriced row becomes $0, re-creating the rows #22 removed.

I did not add an executable rollback, since the project keeps rollbacks as comments. This is a comment-only change to a migration that has not merged yet.

mgcronin (commented):
## Code Review Summary — PR #41 (re-review, cycle 2, head 2b00b91)

Recommendation: merge with follow-ups — 0 blocking findings, 1 new low-severity advisory

This is automated analysis published as a comment, not a human sign-off. The reviewing identity (`mgcronin`) is also the PR author, so this is not an independent approval. A human owns the merge decision.

### Previous blocking finding: fixed, and I checked by running it
The autopilot ceiling now fails closed on unpriced usage (`checkCeiling` in `autopilot-runner.ts`, `projectMonthlyCostForCeiling` in `budget-enforcer.ts`). I ran two mutations in a detached worktree at 2b00b91:
- Removing the `unpricedTokens > 0` throw turns 2 tests red: the all-unpriced start refusal and the mid-run `assertCeiling` refusal.
- Replacing the re-pricing with `r.costCents` turns 1 test red: the MODEL_PRICES re-pricing test.
- With no ceiling set, unpriced usage is still allowed, and a test covers that.
- No other code calls the old `projectMonthlyCost` for a ceiling check.

### Previous advisories
- Fixed: findings 2 to 6, 8, 9, 10 and 11, as the author described.
- Finding 7 was declined and filed as #43. I agree with the reasoning: passing the real provider without failing closed would disable that budget as well.
- The Bedrock Fable row is filed as #42.
- `Refs #25` is in place, and the thinking-format correction is in the PR body.

### New (advisory, low)
- Once MODEL_PRICES is set, the usage summary projection and the ceiling projection count different things. See the inline comment.

### Gate, run locally at 2b00b91
- `pnpm typecheck`: exit 0.
- eslint on the changed server and UI `.ts`/`.tsx` files: exit 0.
- Changed server test files plus `tests/autopilot-runner.test.ts`: 13 files, 194 tests passed.
- UI usage-page tests: 2 files, 14 tests passed.
- I did not run the full `pnpm test` fan-out. CI `api` ran Lint, Typecheck and Test green.

### CI
Everything passes except `api` and `api-outcome`. Inside `api`, only the step "Verify image size budget" failed. That is the known #34, which open PR #38 fixes, and it is not caused by this PR.

### Still open, not blocking
#25 remains open until a live DeepSeek run checks the RAG coverage in the Architecture document and the Phase-1 speed-up.

server/src/lib/autopilot/autopilot-runner.ts
```diff
@@ -39,16 +41,41 @@ export class AutopilotCostCeilingError extends Error {
...
   }
 }
 
+/**
+ * Throw when the month's projection meets `ceilingCents`, or when any of the
+ * month's usage is unpriced (#22 review, PR #41). Returns the projection.
+ */
+async function checkCeiling(projectId: string, ceilingCents: number, now: Date): Promise<number> {
+  const { projectedCents, unpricedTokens } = await projectMonthlyCostForCeiling(projectId, now);
+  if (unpricedTokens > 0) {
```
mgcronin, line 70: Confirmed fixed (previous blocking finding). With this throw removed, `refuses an all-unpriced project that has a ceiling` and `throws mid-run once unpriced usage appears under a ceiling` both go red. Failing closed here, with a message that names MODEL_PRICES, is the right design.

server/src/lib/finops/budget-enforcer.ts
```diff
@@ -254,9 +339,11 @@ export async function summarizeUsage(
     outputTokens,
     totalTokens,
     costCents,
+    unpriced,
     projectedMonthlyCostCents: projectMonthlyFromMtd(mtd.cents, now),
```
mgcronin, line 343: Low, advisory, confidence high. `projectedMonthlyCostCents` in the summary is still built from `mtd.cents`, which leaves out NULL-cost rows. `projectMonthlyCostForCeiling` re-prices those rows once `MODEL_PRICES` is set. So after an admin adds a price, the Projected-month tile keeps showing the earlier rows as unpriced (and understates spend), while the ceiling counts them priced. The two figures can then disagree: autopilot refuses at a ceiling the tile says is far away. It fails safe, so this is not blocking. A follow-up could route the summary projection through the same re-pricing so there is one projection.

Issue #45, opened 2026-09-22
metis-server image cannot start against Postgres: Prisma client is generated for SQLite only


mgcronin:
**The `metis-server` image cannot start against Postgres.** Found while fixing #39.

`Dockerfile.server` runs `prisma generate` with no `DATABASE_URL`, so `server/prisma.config.ts` selects the SQLite schema and the image's generated client is SQLite-only. Measured in an amd64 build of the #39 branch:

```
$ grep activeProvider .../node_modules/.prisma/client/*.js
"activeProvider": "sqlite"
```

`server/src/lib/prisma.ts` picks `@prisma/adapter-pg` for a `postgres://` `DATABASE_URL` (#539), and the client then refuses it at import time:

```
PrismaClientInitializationError: The Driver Adapter `@prisma/adapter-pg`, based on `postgres`,
is not compatible with the provider `sqlite` specified in the Prisma schema.
    at file:///app/server/dist/lib/prisma.js:84:5
```

(observed with `DATABASE_URL=postgres://...` against a `postgres:16-alpine` container). Every documented Postgres deployment is affected: `docker-compose.prod.yml` ("Production targets Postgres (issue #390). Required."), and `deploy/helm/metis/values-prod.yaml` (`DATABASE_URL` postgres via External Secrets). The SQLite default path starts and is what #39's CI smoke step exercises.

The runtime stage also carries a contradictory comment: it deletes Prisma's non-Postgres query compilers because "production deployments use Postgres only", while the only client the image can build is SQLite.

## To do

- [ ] Decide how one image serves both providers (generate both clients to separate output paths and select by scheme, or build per-provider images / a build arg)
- [ ] Extend `scripts/lib/smoke-server-image.mjs` (#39) with a Postgres arm (a `postgres` service container on the `api` job), so this cannot regress silently
- [ ] Correct the query-compiler comment in `Dockerfile.server`

PR #46, opened 2026-09-22
fix: Resolve #39 — metis-server image starts, and CI proves it with a /healthz smoke

### LINKED ISSUE #39: metis-server image built from main does not start (jszip missing, pruned @prisma/debug, musl vs glibc LanceDB)
**The `metis-server` image built from `main` does not start.** Found while measuring #34: the image passes its build and (with #38) its size gate, but `node server/dist/index.js` exits at import time. Nothing in CI runs the image, so every gate is green over it.

Measured on an amd64 build of `main` (45b125c, `docker build --platform linux/amd64 -f Dockerfile.server`), and identically on the #38 branch — none of these are introduced by #38:

| Symptom | Cause |
|---|---|
| `ERR_MODULE_NOT_FOUND: Cannot find package 'jszip' imported from server/dist/lib/connectors/repo/archive-extract.js` — **the server exits on start** | `jszip` is a runtime import of the server but is declared under `devDependencies` in `server/package.json`, so the `--prod` install in the image omits it |
| `import("@prisma/adapter-pg")` / `@prisma/adapter-better-sqlite3` → `Cannot find package '@prisma/debug'` | `Dockerfile.server` deletes `node_modules/.pnpm/@prisma+debug@*`, which `@prisma/driver-adapter-utils` needs. `server/src/lib/prisma.ts` imports both adapters statically, so this alone would also stop the server |
| `import("mysql2/promise")` → `ERR_MODULE_NOT_FOUND` | `Dockerfile.server` deletes `mysql2@*` ("we use postgres only"), but `server/src/lib/connectors/db/drivers/mysql.ts` loads it for the MySQL connector |
| `require("vectordb")` → `Error relocating .../@lancedb/vectordb-linux-x64-gnu/index.node: __register_atfork: symbol not found` | LanceDB ships only a glibc binary; the runtime is `node:20-alpine` (musl). `gcompat` does not cover this symbol |
| `oracledb.initOracleClient({ libDir: "/opt/oracle/instantclient" })` → `DPI-1047 ... Error loading shared library libnnz.so` | the Instant Client dir is not on the loader path (`LD_LIBRARY_PATH` / `ldconfig`), so thick mode cannot load |
| `new (require("better-sqlite3"))(":memory:")` (13.0.2) → segfault | not yet diagnosed; 12.11.1 (used by the Prisma adapter) works. Observed under amd64 emulation on an arm64 host, so re-check natively before acting |

## To do

- [ ] Move `jszip` from `devDependencies` to `dependencies` in `server/package.json`
- [ ] Stop deleting `@prisma+debug@*` and `mysql2@*` in `Dockerfile.server`
- [ ] Decide LanceDB on alpine: a glibc base (`node:20-bookworm-slim`) for the server, or a musl build
- [ ] Put `/opt/oracle/instantclient` on the loader path, or drop the Instant Client (it is ~138 MB of the image — see #34)
- [ ] Add a CI smoke step that starts the built image and waits for `/healthz`, so a DOA image cannot pass `api` again

mgcronin:
Closes #39

## What was wrong

The `metis-server` image built from `main` exited at import time, and every CI gate was green over it because nothing ever started the image. Starting it surfaced one more cause than #39 lists:

| Cause | Fix |
|---|---|
| `jszip` is a runtime import (`archive-extract.ts`) declared under `devDependencies` | moved to `dependencies` |
| `@prisma/debug` deleted; `@prisma/driver-adapter-utils` imports it, and `prisma.ts` loads both adapters statically | no longer deleted |
| `mysql2` deleted; the MySQL connector imports it | no longer deleted |
| **Not in #39:** the boot-time migration guard (`migration-guard.ts`) spawned `pnpm exec prisma migrate deploy`. The runtime image ships no package manager, and the Prisma CLI tree was pruned, so the server refused to start even with the imports fixed | the guard runs the CLI's entry script with `process.execPath` and no shell on any platform. The image keeps the CLI tree (`prisma`, `@prisma/config`, `@prisma/engines`, `effect`, `fast-check`, …) and `prisma.config.ts` |
| LanceDB (`vectordb` 0.21) ships glibc bindings only; the runtime was `node:20-alpine` | runtime base `node:22-trixie-slim` (see below) |
| Oracle Instant Client: `DPI-1047`, because its libraries are not on the loader path | `ld.so.conf.d` entry + `ldconfig` + `libaio1t64` (Trixie renamed it; there is a `libaio.so.1` link). The download moved into its own stage, so `curl`/`unzip` never reach the runtime layer |

## musl vs glibc: glibc

- `vectordb` publishes `linux-{x64,arm64}-gnu` and nothing else (checked its `optionalDependencies`). `gcompat` does not provide `__register_atfork`. Its successor, `@lancedb/lancedb`, does ship musl builds, so going back to alpine would mean migrating the vector store SDK first. That is a code change, and this issue does not cover it.
- **Trixie, not bookworm:** better-sqlite3 13's prebuilt binary needs glibc ≥ 2.38, and bookworm has 2.36. **Node 22, not 20:** that is the repository's `engines.node` floor, and better-sqlite3 13 requires it. The segfault in #39's table did not reproduce on Node 22 / trixie. The smoke probe opens `:memory:` and runs a query.

## Size: over the 900 MB provisional budget, re-measured

Measured on amd64 builds of `main` and this branch (`du`, MiB): **860 → 1,111**. By uncompressed layer bytes: 849 → 1,101 MB.

| Change | MiB |
|---|---|
| glibc base (`node:22-trixie-slim` vs `node:20-alpine`) | +97 |
| Prisma CLI tree kept for the migration guard | +150 |
| `@napi-rs/canvas` ×3, gnu instead of musl | +9 |
| `@prisma/debug`, `mysql2`, `fast-check`, `jszip`, libaio | < 3 |

The `api` job measured **1,067.7 MB** (run 35712009974; the local estimate was 1,065). `DEFAULT_MAX_SERVER_IMAGE_MB` changes from 900 to **1,170**, which is that figure plus about 10%. The #34 figure of 821 MB described an image that did not start. `docs/OPERATIONS.md` > "Container Image Sizes" has the full breakdown and two new reducible contributors: running `prisma migrate deploy` as a separate init step would let the image drop about 150 MiB, and the Instant Client is still about 138 MB. The rationale now appears in all three places the budget is stated (docs, `verify-image-size.mjs`, `ci.yml`).

## The CI smoke step (the main point)

`scripts/lib/smoke-server-image.mjs` runs in the `api` job after the size gate and before cleanup:

1. `docker run` the built image with **its own CMD** and production-legal settings: per-run random `JWT_SECRET`/`VAULT_MASTER_KEY`/`EMBEDDINGS_TOKEN`, `AUTH_MODE=ldap` (production refuses mock), and SQLite in `/tmp`. It does **not** skip the migration guard.
2. Poll `/healthz` until it returns 200 (180 s in CI). The job fails if the container exits or the deadline passes, and the container log is printed.
3. `docker exec` probes for things `/healthz` never touches: LanceDB (connect, create table, count rows), `mysql2/promise`, `better-sqlite3` (`select 1`), `oracledb.initOracleClient` in thick mode, and whether the runtime user can write `~/.metis-sessions`. If any probe fails, the job fails.

`docker` is always called with an argument array, never a shell string. The container is always removed.

**It fails against the broken image.** Local amd64 build of `origin/main` (77f23ee):

```
==> Starting metis-server:i39-main-amd64 as metis-smoke-93416-... (127.0.0.1:14391 -> 4000)
FAIL: metis-server:i39-main-amd64 exited before serving /healthz (exit code 1)
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@prisma/debug' imported from
  /app/node_modules/.pnpm/@prisma+driver-adapter-utils@7.8.0/.../dist/index.mjs
smoke exit=1
```

The same build of this branch:

```
OK: /healthz returned 200
OK: module probe lancedb
OK: module probe mysql2
OK: module probe better-sqlite3
OK: module probe oracle-thick
OK: module probe home-writable   (added after self-review; see below)
PASS: metis-server:i39-head2-amd64 starts, serves /healthz and loads 5 runtime module(s)
```

**In CI** (run 35715278015, `api` job, head 93e2174):

```
metis-server:ci-35715278015          1067.8         1170
OK: /healthz returned 200
OK: module probe lancedb
OK: module probe mysql2
OK: module probe better-sqlite3
OK: module probe home-writable
OK: module probe oracle-thick
PASS: metis-server:ci-35715278015 starts, serves /healthz and loads 5 runtime module(s)
```

## Tests and mutation proofs

- `scripts/lib/smoke-server-image.test.mjs`: 16 arms covering args, the pass path, container exit, inspect failure, the deadline, "no response", a failed probe, `docker run` failure, no docker (exit 2), and that secrets are never logged.
- `scripts/lib/image-build-wiring.test.mjs`: the Dockerfile keeps the Prisma CLI, `@prisma/debug`, `mysql2` and `prisma.config.ts`. The runner base is not alpine. The Instant Client is on the loader path. `jszip` is a runtime dependency. `ci.yml` runs the smoke step against this run's tag, between the server build and cleanup, with no `continue-on-error` and no `|| true`.
- `server/tests/migration-guard.test.ts`: the guard spawns `process.execPath <prisma/build/index.js> migrate deploy`, resolves the real CLI, fails loudly when the CLI is missing, and uses no shell on win32 or POSIX.

Mutations, each seen red and then restored:

| Mutation | Went red |
|---|---|
| `Dockerfile.server`, `ci.yml`, `server/package.json` reverted to `origin/main` | 9 wiring tests |
| smoke step gets `\|\| true` / `continue-on-error: true` | "is not allowed to fail quietly" (each) |
| smoke ignores container exit (`if (false)`) | 2 (the #39 shape, inspect failure) |
| smoke ignores failed probes | the failed-probe arm |
| smoke treats any HTTP status as healthy | the deadline arm |
| smoke ignores the deadline | 2 |
| smoke drops the `rm --force` in `finally` | 2 |
| `migration-guard.ts` reverted to `origin/main` | 4 |
| `useradd` without `--create-home` | the home-directory wiring test |
| guard restores `shell: process.platform === "win32"` | the win32 arm |

## Coverage

- `scripts/lib/smoke-server-image.mjs` (new): 100% statements, 98% branches, 100% functions, 100% lines.
- `server/src/lib/db/migration-guard.ts`: 100% statements, 80.95% branches, 100% lines.

## Gate (run locally before push)

`pnpm lint` 0, `pnpm typecheck` 0, `pnpm test` 0 (server 1,099 files passed; scripts 57 files / 2,317 tests; ui 350 files). `pnpm agents:verify`, `pnpm skills:verify` and `pnpm changelog:verify --base origin/main` all pass.

## Security

- No new dependency: `jszip` 3.10.1 moved from dev to prod at the same version.
- The smoke script uses `spawnSync` with argv arrays and no shell. Its secrets are random per run, never printed, and gone when the container is removed. The port is published on loopback only.
- The migration guard now spawns no shell on any platform (Windows used to get one), with fixed argv.
- The image still runs as `USER metis`. The Oracle download keeps its existing HTTPS URL and behaviour. It moved stages and was not otherwise changed.

- **Found in self-review and fixed here:** alpine's `adduser --system` created `/home/metis`. Debian's `useradd --system` does not. On the first glibc build, the server's `~/.metis-sessions` and `~/.metis/auth.json` writes would have failed with `EACCES` even though `/healthz` was green. The new `home-writable` probe **failed against that build** (`EACCES: permission denied, mkdir '/home/metis/.metis-sessions/smoke'`, smoke exit 1). With `useradd --create-home` it passes, and a wiring test pins `--create-home` (it went red when the flag was removed).

- **#45 (filed):** the image cannot start against **Postgres**. `prisma generate` runs with no `DATABASE_URL`, so the only generated client is SQLite (`"activeProvider": "sqlite"` in the image), and `prisma.ts` then fails with `adapter-pg ... not compatible with the provider sqlite`. This affects `docker-compose.prod.yml` and the helm prod values. This PR's smoke covers SQLite only, and #45 includes adding a Postgres arm.

## Deferred

- Postgres startup (#45).
- Shrinking the image: a separate migrate init step (about 150 MiB), an Instant Client build arg (about 138 MB), and migrating to `@lancedb/lancedb` to reopen alpine.


mgcronin (commented):
Self-review of the final diff (93e2174). This is the author's pass, not an approval.

- **Found and fixed during review:** the switch from alpine to Debian dropped the runtime user's home directory, because `useradd --system` creates none. The fix is `--create-home`. A new `home-writable` smoke probe catches it: it failed against the earlier build with EACCES.
- **Checked and left as is:** `createRequire(import.meta.url).resolve("prisma/build/index.js")` resolves both in a dev checkout (the unit test asserts the file exists) and in the image (the migration guard ran at boot in every smoke).
- **Checked and left as is:** the smoke passes secrets on the `docker run` argv. They are random per run, never printed, loopback-only, and removed with the container.
- **Out of scope, filed as #45:** Postgres `DATABASE_URL` fails at import because the only generated client is SQLite. This PR's smoke covers the SQLite default only.
- **Size:** 1,067.8 MB in CI against a 1,170 budget. The two largest reducible items (a separate migrate step, and the Instant Client) are listed in OPERATIONS.md, not done here.

scripts/lib/smoke-server-image.mjs
```diff
@@ -0,0 +1,318 @@
...
+ * @type {ReadonlyArray<{ name: string, code: string }>}
+ */
+export const MODULE_PROBES = Object.freeze([
+  {
+    // The RAG vector store (server/src/lib/rag/vector-store.ts). Its native
+    // binding ships for glibc only; a musl runtime fails here, not at /healthz.
+    name: "lancedb",
+    code: [
+      'const lancedb = await import("vectordb");',
+      'const db = await lancedb.connect("/tmp/metis-smoke-lance");',
```
mgcronin, line 53: **Medium, advisory (high confidence): this probe writes to `/tmp`, so it can pass while RAG writes fail at the path the server actually uses.** `vector-store.ts:1256` defaults to `process.env.LANCEDB_PATH ?? path.resolve(process.cwd(), "data", "lancedb")`, and the CMD's cwd is `/app`. `/app` is `root:root 755` in this image, and the process runs as `metis`. I ran the same `connect` + `createTable` against `/app/data/lancedb` as uid 1001 in the PR-head image and got `Unable to created lance dataset at /app/data/lancedb: Permission denied (os error 13)`. This is not a regression: `/app` was root-owned on alpine too. Deployments that set `LANCEDB_PATH` or mount a volume are fine. But the smoke's comment says a module that only fails on first use should be a red build, and this is that case. Suggestion: either create and chown `/app/data` in the runner stage, or point the probe at the server's default root. Otherwise, say in the probe that it tests loading and not the default path. A follow-up issue is fine.

scripts/lib/smoke-server-image.mjs
```diff
@@ -0,0 +1,318 @@
...
+    `VAULT_MASTER_KEY=${vaultKey}`,
+    // Production refuses the mock provider (#682). LDAP is the one real provider
+    // that needs no reachable IdP to boot: it only connects on a login.
+    "--env",
+    "AUTH_MODE=ldap",
+    "--env",
+    "LDAP_URL=ldap://127.0.0.1:9",
+    // SQLite in the container's own /tmp, so the smoke needs no database service.
+    "--env",
+    "DATABASE_URL=file:/tmp/metis-smoke.db",
```
mgcronin, line 143: **Medium, advisory (tracked in #45): the smoke only covers SQLite.** `docker-compose.prod.yml` and the Helm prod values use Postgres, which still fails at startup because the image generates only a SQLite client. So a green `api` job means the image boots on SQLite, not in the documented prod topology. The memory file says this correctly. It already failed before this PR and is outside the scope #39 listed, so it does not block. When #45 lands, please add a second smoke run with a Postgres service, so that fix is gated in CI the same way this one is.

.github/workflows/ci.yml
```diff
@@ -250,6 +249,20 @@ jobs:
...
+      # devDependency, a pruned `@prisma/debug`, a glibc-only LanceDB binding on a
+      # musl base) while every gate here was green, because nothing ever started it.
+      # This starts the built image with its own CMD, polls /healthz until 200, then
+      # loads inside the running container the native/lazy modules /healthz never
+      # touches (LanceDB, mysql2, better-sqlite3, Oracle thick mode) and checks the
+      # runtime user can write its home directory. A container that exits, a
+      # /healthz that never answers, or a probe that fails fails the job, with the
+      # container log printed.
+      - name: Smoke-test metis-server (starts, serves /healthz)
+        if: ${{ !startsWith(github.head_ref, 'dependabot/') }}
```
mgcronin, line 263: **Low, advisory: Dependabot PRs skip the smoke.** This matches the existing policy: images are not built for dependabot PRs, so there would be nothing to smoke. But a dependency bump, for example a new major version that moves a runtime import or changes a native binding, is the most likely way the next DOA image appears. It would then first go red on `main`, after merge. Consider running the server build and smoke (without the ui/embeddings/sql-lineage builds) on dependabot PRs that touch `server/package.json` or `pnpm-lock.yaml`. A follow-up is fine.

server/src/lib/db/migration-guard.ts
```diff
@@ -50,6 +53,15 @@ function resolveServerRoot(): string {
   return path.resolve(here, "..", "..", "..");
 }
 
+/**
+ * The Prisma CLI's entry script, resolved from this package the way `server/dist`
+ * resolves every other dependency. `prisma` is a server `dependency`, so this is
+ * present in a dev checkout and in the production image alike.
+ */
+function resolvePrismaCliDefault(): string {
+  return createRequire(import.meta.url).resolve("prisma/build/index.js");
```
mgcronin, line 62: **Low, informational: this relies on a deep path inside `prisma`.** `prisma/build/index.js` resolves today; I checked with prisma 7.8.0 from `server/`. It depends on prisma's `exports` map keeping that path reachable, and a Prisma upgrade could drop it. The error message is clear and the image smoke would catch it, so no change is needed. A one-line note in the JSDoc that this path is not a public entry point would help whoever does the next Prisma bump.

server/src/lib/db/migration-guard.ts
```diff
@@ -60,20 +72,26 @@ export async function ensureSchemaUpToDate(
...
     cwd,
     env,
     stdio: ["ignore", "pipe", "pipe"],
     encoding: "utf-8",
-    // On Windows `pnpm` is a `.cmd` shim; spawnSync cannot resolve it without a
-    // shell and fails with ENOENT, so the server refuses to start on every
-    // Windows dev box. Use a shell on win32 only — POSIX stays shell-free (more
-    // secure). The argv is hardcoded, so there is no shell-injection surface.
-    shell: process.platform === "win32",
+    shell: false,
```
mgcronin, line 94: **Low, advisory: the Windows behaviour change is only covered by a mock.** The Windows `shell: true` special case is removed. That is correct in principle: `node <script.js>` needs no `.cmd` shim, and this is a real security improvement. But the only test fakes `process.platform` with a stubbed `spawn`, and the `windows` CI job is skipped unless `vars.WINDOWS_CI_ENABLED` is set. It was skipped on this run. One manual `pnpm dev` start on a Windows box, or a run with the variable set, would close this.

Dockerfile.server
```diff
@@ -322,8 +321,29 @@ RUN set -eux; \
...
+RUN apt-get update \
+ && apt-get install -y --no-install-recommends ca-certificates curl unzip \
+ && rm -rf /var/lib/apt/lists/* \
+ && mkdir -p /opt/oracle \
+ && if [ "$TARGETARCH" = "arm64" ]; then \
+      OIC_URL="https://download.oracle.com/otn_software/linux/instantclient/2390000/instantclient-basiclite-linux.arm64-23.9.0.25.07.zip"; \
+    else \
+      OIC_URL="https://download.oracle.com/otn_software/linux/instantclient/2390000/instantclient-basiclite-linux.x64-23.9.0.25.07.zip"; \
+    fi \
+ && curl -fsSL "$OIC_URL" -o /tmp/ic.zip \
```
mgcronin, line 339: **Low, informational (A08; not new in this PR): the Instant Client zip is downloaded without a checksum.** The same `curl` is only moved into its own stage here. Pinning a SHA-256 per arch and checking it with `sha256sum -c` would protect this layer if the CDN is compromised or the file is quietly swapped. Moving it into a separate stage is itself an improvement, because curl and unzip no longer reach the runtime image.

Dockerfile.server
```diff
@@ -129,46 +141,35 @@ RUN pnpm install --prod --frozen-lockfile \
...
  && rm -rf /root/.cache /root/.npm /root/.local /tmp/* /usr/local/share/pnpm/store
 
-# Post-install prune: drop transitive dev-time fat that @prisma/client lists
-# as optional peer deps (typescript, prisma CLI, effect from @prisma/config).
-# These are NOT needed at runtime — the Prisma client has already been
-# generated above. Keeping them adds ~110 MB for nothing.
+# Post-install prune: the TypeScript compiler is a peer that nothing at runtime
+# runs (prisma.config.ts is loaded by the Prisma CLI through jiti, not tsc).
+#
+# #39 — the Prisma CLI and its tree (@prisma/config, @prisma/engines, effect, c12,
```
mgcronin, line 147: **Informational: this keeps the full Prisma CLI tree in the runtime image.** That includes `@prisma/studio-core`, `@prisma/dev`, `@electric-sql/pglite` and `react-dom`. It is about +150 MiB and extra runtime attack surface, all so the server can migrate itself at boot. The trade-off is reasonable for #39 and OPERATIONS.md records running migrations as a separate init step as the way to get this back. Agreed it should be a follow-up and not part of this PR.

docs/OPERATIONS.md
```diff
@@ -394,37 +394,56 @@ What is left, measured in the amd64 image (MiB, from `du`):
 
 | Component | Size | Notes |
 |---|---|---|
-| Oracle Instant Client (basiclite 23.9, `.so` only) | ~138 | Intended for thick mode (pre-12c password verifiers); does not load today (#39) |
+| Oracle Instant Client (basiclite 23.9, `.so` only) | ~138 | Thick mode (pre-12c password verifiers); did not load in this image — loads since #39 |
 | Node.js 20 runtime (`/usr/local`) | ~104 | The `node` binary alone is 97 |
```
mgcronin, line 398: nit: this row still says "Node.js 20 runtime". The runtime is now Node 22 on trixie-slim. The table is labelled as the #34 breakdown, so either add "(at #34)" or update the row, so it is not read as the current state.

scripts/lib/smoke-server-image.mjs
```diff
@@ -0,0 +1,318 @@
...
+    "run",
+    "--detach",
+    "--name",
+    name,
+    "--publish",
+    `127.0.0.1:${port}:${CONTAINER_PORT}`,
+    // Real secrets, generated per run: production refuses the published dev JWT
+    // key (#1057) and requires a vault key.
+    "--env",
+    `JWT_SECRET=${jwtSecret}`,
```
mgcronin, line 132: nit / informational: the secrets are passed as `--env KEY=value` on docker's argv, so anyone who can run `ps` on the runner can see them while the step runs. They are random for each run and exist only on an ephemeral runner, so the risk is negligible. `--env JWT_SECRET` (no value), with the value in the spawned process's `env`, would keep them off argv if this script is ever reused on a shared host.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking, 9 non-blocking findings

This is an automated review posted as a comment. It is not a human sign-off. The reviewing `gh` identity (mgcronin) is also the PR author, so GitHub cannot record an approval or a change request from it. The merge decision is yours.

## Code Review Summary — PR #46 (Closes #39)

### Requirements: PASS (1 known gap tracked as #45)
- [x] `jszip` moved to `dependencies` (server/package.json; the lockfile diff is only that move)
- [x] `@prisma/debug`, `mysql2` and `fast-check` are no longer pruned
- [x] musl vs glibc decided with reasons: `node:22-trixie-slim`, because `vectordb` has no musl build. The +97 / +150 / +9 MiB costs are itemised in docs/OPERATIONS.md, and the budget is set to the CI measurement plus 10%, not raised until CI passed
- [x] Oracle Instant Client is on the loader path (ld.so.conf.d + libaio), and thick mode loads
- [x] A CI smoke step starts the image, polls /healthz and fails the job. It is wired into the `api` job between the size gate and cleanup, with no `continue-on-error`
- [x] Found and fixed a cause #39 did not list: the migration guard spawned `pnpm`, which the image does not have
- [x] Changelog fragment `.changes/unreleased/39-server-image-starts.md` is present and `changelog:verify` passes. Memory file marked resolved; `agents:verify` passes
- [ ] Gap: the smoke covers SQLite only. The prod topology (Postgres) still cannot start. This was already broken before this PR and is tracked in #45

### Claims I checked myself
- **The smoke fails on the broken image.** I built `origin/main` (e347c2f) natively on arm64 and ran this PR's smoke script against it. It exited 1 with `FAIL: ... exited before serving /healthz (exit code 1)` and printed the container log: `ERR_MODULE_NOT_FOUND: Cannot find package 'jszip'`. The author saw `@prisma/debug` instead, but the result is the same: a red build.
- **The fix passes, on arm64 as well.** I built PR head 93e2174 natively on arm64 and the smoke passed: /healthz 200 and all 5 probes (lancedb, mysql2, better-sqlite3, home-writable, oracle-thick). This covers the arm64 Oracle and libaio path, which CI (amd64) never runs.
- **Migrations actually run.** I started the image by hand. The log shows `Applying pending Prisma migrations` and then `database is up-to-date`. The SQLite file was created (3.1 MB), and `/readyz` reports `database: ok`. Because the guard runs before `listen()`, a /healthz 200 does prove that migrations ran.
- **Mutations, each reverted afterwards:**
  - `migration-guard.ts` put back to origin/main: 4 tests failed.
  - Smoke treating any HTTP status as healthy: 1 failed.
  - Container removal dropped from `finally`: 2 failed.
  - Probe failures ignored: 1 failed.
  - `Dockerfile.server` put back to origin/main: 7 wiring tests failed.
  - The author's falsifiability claims hold.
- **CI.** Run 35715278015: the full check set is present with 0 pending. `api`, `api-outcome` and `postgres-adapter` passed. `windows` was skipped because it only runs when `vars.WINDOWS_CI_ENABLED` is set; `e2e` was skipped. The smoke step's log shows PASS in about 7s.
- **Local gate, scoped to this PR.** Changed-file tests: 76 passed in scripts/ and 8 in server/. ESLint on the changed server files is clean. The changed `scripts/lib` files are excluded by the lint config, so ESLint only reported "file ignored" warnings for them, and CI `lint` passed. `tsc --noEmit` on server passes. I did not re-run the full `pnpm test` fan-out locally; CI ran it.

### Security: CLEAN (informational only)
- No new dependency. `docker` and the Prisma CLI are called with argument arrays and no shell. The Windows shell is removed.
- Informational items are in the inline comments: the Oracle zip is not checksum-verified (this was already the case), the smoke's secrets appear on docker's argv, and the Prisma CLI tree (studio-core, pglite, react-dom) is now in the runtime attack surface.

### Tests: SUFFICIENT
The smoke script is at 100/98/100/100 (statements/branches/functions/lines) per the author; I did not re-measure. `migration-guard.ts` is at 80.95% branches. The mutations above went red.

### Documentation: UP TO DATE (1 nit)
OPERATIONS.md has the measured figures and the causes. One row in the breakdown still says "Node.js 20".

### Scope of this review
- Read all 16 changed files, and the full smoke script and migration guard.
- Built both images and ran the smoke against each.
- Ran the container by hand to check that migrations ran and that `/readyz` works.
- Did not exercise Postgres (#45), a real LanceDB ingest, or Windows.

PR #47, opened 2026-09-22
fix: Resolve #40, #35, #24 — multi-call tool prompt, 4xx body-parser errors, METIS-only restart sweep

### LINKED ISSUE #24: scripts/restart.sh kills every MCP server on the machine, not just METIS's
**`scripts/restart.sh` sends SIGTERM to any process whose command line matches `modelcontextprotocol|@modelcontextprotocol|mcp-server-`.** That includes MCP servers owned by other tools — editors and coding agents running on the same machine lose their MCP connections whenever METIS is restarted.

Observed: running the script terminated eight processes, including an unrelated coding agent's GitHub and Playwright MCP servers, while no METIS process was running.

## Acceptance criteria

- [ ] The orphan sweep only targets processes METIS started — e.g. descendants of the dev server's process tree, or processes carrying a METIS-specific marker (environment variable or argv tag set when spawning sidecars)
- [ ] Running the script with no METIS processes running terminates nothing

### LINKED ISSUE #35: Oversized and other body-parser client errors still return 500 (413/415 expected)

### LINKED ISSUE #40: Agent-loop prompt still says "Call tools one at a time" after #15 made multi-call replies work
Follow-up from the review of PR #37 (#15).

`server/src/lib/analysis/agent-loop.ts` still tells the model "Call tools one at a time. After receiving a tool result, you may call another tool or provide your final answer." in both tool-protocol prompt blocks (around lines 586 and 609 on the PR #37 branch). Since #15, the loop runs every call in one reply (up to `MAX_TOOL_CALLS_PER_REPLY` = 8), in order, and returns all results on the next turn. The prompt now says less than the loop can do, and a model that obeys it spends one turn per search.

PR #37 left the wording alone on purpose. The text is part of the cached system-prompt prefix, and changing it in the same PR would have mixed a parsing fix with a cache-invalidating prompt change.

## Acceptance criteria

- [ ] The prompt says the model may request several tools in one reply, and names the accepted forms and the per-reply cap
- [ ] The effect on the prompt-cache prefix is measured or stated (the prefix changes once)
- [ ] Existing prompt snapshot tests updated; no other prompt text changes

mgcronin:
Three small follow-ups in one PR. Each has its own tests, and each test was seen to fail against the unfixed code.

Closes #40
Closes #35
Closes #24

## #40: the tool prompt no longer says "one at a time"

- `server/src/lib/analysis/agent-loop.ts`: the `compact` and `full` manifests both had the sentence *"Call tools one at a time…"*. It is now replaced by one shared `multiCallInstruction()`. That sentence says a reply may request several tools, as consecutive JSON objects or as a JSON array, at most `MAX_TOOL_CALLS_PER_REPLY` (8) per reply, run in order, with all results returned on the next turn. The cap comes from the constant, so the prompt and the loop can't drift apart.
- **Found along the way: the issue named two blocks, but there is a third.** `formatToolSchemas`, the `"schema"` format and the **default** cached prefix since #398, also said *"Call one tool at a time"*. Updating only the two named blocks would have left the prompt most runs actually use unchanged. So it is fixed here too, and only that sentence changed.
- **Prompt-cache prefix: measured, and it changes once.** With a one-tool fixture, using the repo's `estimateCachedPrefixTokens`:

  | format | chars before → after | est. tokens before → after |
  |---|---|---|
  | schema (default) | 3406 → 3578 | 852 → 895 (+43) |
  | compact | 3207 → 3364 | 802 → 841 (+39) |
  | full | 3134 → 3291 | 784 → 823 (+39) |

  The first analysis after upgrading writes a new cache entry. After that the prefix is byte-stable again. The Sonnet cache-floor tests in `agent-loop-caching.test.ts` still pass.
- "Existing prompt snapshot tests": none existed for this text (grep found no snapshots that contained it). `tool-protocol-prompt.test.ts` now pins the full protocol block of all three formats byte for byte, so any other prompt-text change fails it. It also checks that both advertised forms parse into two calls with the shipped `parseToolCalls`.

## #35: body-parser client errors return 4xx, not 500

`server/src/middleware/error-handler.ts` gets a `BODY_PARSER_CLIENT_ERRORS` table, placed after the #21 `INVALID_JSON` branch:

| body-parser `type` | status | code |
|---|---|---|
| `entity.too.large` | 413 | `PAYLOAD_TOO_LARGE` (the code the codebase already uses) |
| `parameters.too.many` | 413 | `TOO_MANY_PARAMETERS` |
| `encoding.unsupported` | 415 | `UNSUPPORTED_CONTENT_ENCODING` |
| `charset.unsupported` | 415 | `UNSUPPORTED_CHARSET` |
| `request.aborted` | 400 | `REQUEST_ABORTED` |
| `request.size.invalid` | 400 | `REQUEST_SIZE_MISMATCH` |
| `querystring.parse.rangeError` | 400 | `FORM_BODY_TOO_DEEP` |

- An error matches only when its `type` is in the table **and** its `status` equals the expected status. An unrelated error that happens to carry `type: "entity.too.large"` stays a 500. The lookup uses `Object.hasOwn`, so prototype keys never match.
- `err.message` is never forwarded, because body-parser's messages quote the client's `Content-Encoding` or charset. The log line records type, status, method, path and the numeric `limit`/`length`, but never the body.
- The issue listed five types. `parameters.too.many` and `querystring.parse.rangeError` are the two other client errors body-parser 2.2.2 raises (from `urlencoded`, which `createApp` mounts), so they are included. `entity.verify.failed` (403) was left alone: only our own `verify` hook can raise it, and ours never throws.
- Tested through `createApp` in `server/tests/request-body-client-errors-route.test.ts`: an 11 MB body sent past the real 10 MB `express.json()` gets **413 `PAYLOAD_TOO_LARGE`**, and an unknown `Content-Encoding` gets 415.

## #24: `restart.sh` stops only what METIS started

What was wrong: `stop_stack` SIGTERMed everything matching `modelcontextprotocol|@modelcontextprotocol|mcp-server-`. Its dev patterns (`next dev`, `tsx … src/index.ts`) and port holders on 3000/4000 also matched other projects, so the issue's second acceptance criterion ("with no METIS processes running, terminates nothing") could not hold even with the MCP pattern removed.

- **Candidates** (port holders plus the existing dev argv patterns) count only when their working directory is inside this checkout (`/proc/<pid>/cwd`, or `lsof -d cwd` on macOS). Everything **descended** from such a process is stopped with it, so MCP sidecars are included while the server lives.
- **Orphans**: `stdio-transport.ts` now stamps `METIS_SIDECAR_OWNER` on every sidecar. `restart.sh` exports a per-checkout value (`metis-<cksum of the checkout path>`) into the `pnpm dev` it launches, and sweeps processes whose **environment** carries that exact value: `/proc/<pid>/environ` on Linux, or on macOS the part of `ps -E` output that comes after the argv. A command line that merely mentions the tag does not match. The marker is spread last, so a configured MCP server env can't remove or spoof it. The value is a hash, not a secret, and not the path. A server started without `restart.sh` marks sidecars with the bare `metis`, which no sweep matches; they are still stopped as descendants while their server runs.
- Found while testing: a forked subshell reports the environment its process was **exec'd** with. A script started with `METIS_SIDECAR_OWNER` set, where an `unset` alone would not help, would find and SIGTERM its own sweep subshells. The script now re-execs itself without the variable. There is a test for this.
- A process on a METIS port that did not start from this checkout (another project, or Docker's port proxy) is no longer killed. The existing "port still bound" warning now names it.
- `restart.ps1` (Windows): the MCP argv pattern is removed. `Stop-Tree` already runs `taskkill /T`, so sidecars still go down with their server. The Windows dev-pattern and port matching is **not** narrowed (see Deferred).

**Isolation: nothing was run against the live machine.** `scripts/lib/restart-sh-sweep.test.mjs` runs a *copy* of the script whose repo root is an empty temp dir, on free ports, with `BASH_ENV` defining a `kill` function. Bash prefers the function over the builtin, so `kill -0` passes through and every real signal is only recorded. Running the unfixed script under that harness recorded two real MCP-server PIDs on this machine, and nothing was signalled.

## Mutations: each one made the named tests fail, then was reverted

| # | Mutation | Failed |
|---|---|---|
| #40 | pre-fix code (the original sentence in all three blocks) | 6 of 7 in `tool-protocol-prompt.test.ts` |
| #40 | hard-code `at most 5` in the schema block | schema exact-text test + schema `it.each` |
| #40 | drop "of them" from the shared sentence | compact + full exact-text tests |
| #35 | pre-fix code (no branch) | all 10 new tests (8 unit, 2 through `createApp`) |
| #35 | remove the `status !== mapped.statusCode` guard | "type matches but status does not" test |
| #35 | map `entity.too.large` to 400 | 3 tests, including the `createApp` 413 test |
| #35 | forward `err.message` | 5 no-echo tests |
| #24 | **unfixed `restart.sh` from `origin/main`** | all 5 POSIX tests (it recorded PIDs 53289 and 53256, real MCP servers on this machine, plus the decoys) |
| #24 | `runs_in_repo` always true | 5 tests |
| #24 | `descendants_of` returns nothing | whole-tree test |
| #24 | owner-tag sweep disabled | orphan test |
| #24 | macOS tag match over argv+env instead of env only | "signals nothing" test + own-subshell test |
| #24 | `unset` instead of re-exec | own-subshell test |
| #24 | `stdio-transport`: marker not stamped | 3 marker tests |
| #24 | `stdio-transport`: marker spread before configured env | spoof test |
| #24 | `stdio-transport`: exported tag ignored | 2 marker tests |

## Gates (run locally in the worktree, per CLAUDE.md)

- `pnpm lint`: exit 0 (NUL check, identifier scan covering the new files once staged, license metadata, CLA, `eslint . --max-warnings=0`). `shellcheck scripts/restart.sh`: clean.
- `pnpm typecheck`: exit 0.
- `pnpm test`: exit 0. server 1106 files / 17291 tests passed, ui 3249, scripts 2298, shared 698, embeddings-svc 147, copilot-svc 78, ui-kit 7.
- `pnpm changelog:verify`: 3 fragments contributed and all parse.

## Coverage (lines, v8, targeted suites)

| file | before | after |
|---|---|---|
| `server/src/middleware/error-handler.ts` | 95.08% | 96.00% (new branch fully covered; uncovered lines are the pre-existing 500 fallback) |
| `server/src/lib/analysis/agent-loop.ts` | 95.19% | 95.20% (changed functions fully covered) |
| `server/src/lib/mcp/stdio-transport.ts` | 85.95% | 86.06% |
| `scripts/restart.sh` | n/a | v8 cannot measure bash. Every new function runs in the 5 subprocess tests. The macOS branches ran locally; the Linux `/proc` branches run on the ubuntu CI job. |

## Security (OWASP pass on the diff)

- No new dependencies, so no audit delta.
- A03 injection: `restart.sh` passes only PIDs taken from `ps` or `/proc` to `kill`. The tag is `metis-<digits>`, so it is safe inside the awk string. `REPO_ROOT` is quoted in the `case` pattern, so glob characters in the path are literal.
- A09 information leak: no parser message or body fragment reaches the client or the log.
- The sidecar env gains one non-secret variable. SEC-4's allow-list of inherited parent variables is unchanged.

## Deferred

- `restart.ps1` still matches `next dev`/`tsx`/pnpm patterns and port holders machine-wide on Windows. `Win32_Process` exposes no working directory, so the cwd rule doesn't port directly. This can't be exercised here (no `pwsh`), so it was left out rather than shipped untested.
- Filed as #48. No new issue was filed for the bundle itself, because #40, #35 and #24 already are the record.


scripts/restart.sh
```diff
@@ -100,47 +116,156 @@ pids_matching() {
...
-  # 1. Collect candidate PIDs.
+# ----------------------------------------------------------------------------
+# runs_in_repo <pid> — true when the process's cwd is inside this checkout
+# ----------------------------------------------------------------------------
+runs_in_repo() {
+  local cwd
+  cwd="$(process_cwd "$1")"
+  [ -n "$cwd" ] || return 1
+  case "$cwd/" in
+    "$REPO_ROOT"/*) return 0 ;;
```
mgcronin, line 138: **Medium, non-blocking, high confidence (verified).** `"$REPO_ROOT"/*` also matches every nested git worktree under `$REPO_ROOT/.claude/worktrees/agent-*`, which CLAUDE.md requires parallel implementers to use. I copied this script into a temp repo and started `node ... next dev` with cwd `repo/.claude/worktrees/agent-x`. With the recording kill shim, `--stop-only` from the main checkout sent TERM and then KILL to it. This is no worse than before, since the old sweep killed it too. But it contradicts the fragment's "dev processes running from the same checkout", and it will take down another agent's dev server or tsx server mid-run. Suggested fix: exclude `"$REPO_ROOT"/.claude/worktrees/*` before the accept arm, or, more generally, compare `git -C "$cwd" rev-parse --show-toplevel` to `$REPO_ROOT`. Add a test case that uses a nested directory.

scripts/restart.sh
```diff
@@ -100,47 +116,156 @@ pids_matching() {
...
   # Use a space-delimited string for seen tracking (bash 3.2 compatible —
   # macOS ships bash 3.2 which does not support declare -A).
   # Guard the array expansion with a length check: in bash < 4.4, expanding
   # an empty array with [@] under set -u triggers "unbound variable".
-  local self_pid=$$
-  local parent_pid=$PPID
-  local pids=()
   local seen_pids=" "
+  local p
+  while IFS= read -r p; do [ -n "$p" ] && seen_pids="$seen_pids$p "; done < <(ancestors_of "$$")
```
mgcronin, line 249: **Low, non-blocking, medium confidence (by reading, not run).** The exclusion set is `$$` plus its ancestors. If an ancestor is itself a root, for example an agent shell `bash -c 'cd metis && pnpm dev; ./scripts/restart.sh'` whose cwd is in the repo and whose argv matches `metis.*pnpm dev`, then `descendants_of` at L238 also returns the script's own subshells, the `$(...)` and `<(...)` workers. Only `$$` is filtered, not those subshells, so the script can signal its own workers. It is an edge case, and most of those workers have exited before `kill_pids` runs. Possible fix: also drop descendants of `$$` (for example, compute `descendants_of $$` and add them to `seen_pids`).

scripts/restart.ps1
```diff
@@ -86,8 +87,7 @@ function Stop-Stack {
     'next dev',
     'pnpm.*--filter ./server.*dev',
     'pnpm.*--filter ./ui.*dev',
-    'metis.*pnpm dev',
-    'modelcontextprotocol|@modelcontextprotocol|mcp-server-'
+    'metis.*pnpm dev'
```
mgcronin, line 90: **Low, non-blocking.** Dropping the MCP pattern is correct. But on Windows, a sidecar orphaned by a server that already died is now never swept: `taskkill /T` only covers trees that are still attached, and there is no owner-tag search on this side. #48 lists the over-broad dev and port matching but not this lost orphan path. Please add it to #48 so the Windows follow-up restores orphan cleanup through the owner tag or `ParentProcessId` chains.

scripts/restart.sh
```diff
@@ -100,47 +116,156 @@ pids_matching() {
...
+# ----------------------------------------------------------------------------
+# pids_with_owner_tag — processes whose ENVIRONMENT carries this checkout's
+# METIS_SIDECAR_OWNER tag (argv is never consulted, so a command line that
+# merely mentions the tag does not match).
+# ----------------------------------------------------------------------------
+pids_with_owner_tag() {
+  local marker="METIS_SIDECAR_OWNER=$SIDECAR_OWNER"
+  if [ -r /proc/self/environ ]; then
+    local f pid kv
+    for f in /proc/[0-9]*/environ; do
```
mgcronin, line 152: **Low (performance), non-blocking.** This loop does a bash `read -d ''` for every environment variable of every readable process. On a busy Linux dev box (1-2k processes with around 50 vars each) that is tens of thousands of builtin reads, which could add seconds to every restart. I did not measure it. `grep -qzxF "$marker" "$f" 2>/dev/null && echo "$pid"` gives the same exact-entry match (`-z` NUL-delimited, `-x` whole entry) in C. Note: the macOS branch was the one the implementer ran, and the Linux branch passed on CI.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking, 4 non-blocking findings

This is an automated analysis published as a COMMENT, not a human sign-off: the reviewing gh identity (mgcronin) is also the PR author, so GitHub cannot record an independent approval. The merge decision stays with a human.

## Code Review Summary — PR #47 (closes #40, #35, #24)

### Requirements: PASS
- #40: all three tool-protocol blocks (schema, compact, full) now advertise multi-call replies, both accepted forms (consecutive objects, JSON array) and the cap interpolated from `MAX_TOOL_CALLS_PER_REPLY`; cache-prefix change stated in the fragment. Catching the third (`formatToolSchemas`) block the issue did not name was right. The rephrasing in that block ("Call a tool by", "After the tool results") is part of the same sentence and in scope. The cap claim matches the loop, which tells the model when calls past 8 were dropped (agent-loop.ts ~1443).
- #35: every mapped `type`/status pair checked against the installed body-parser 2.2.2 and raw-body 3.0.2 sources (413 entity.too.large / parameters.too.many, 415 encoding/charset, 400 request.aborted / request.size.invalid / querystring.parse.rangeError). Messages are fixed strings, the body and parser message are neither returned nor logged, and type+status must both match. A `createApp` test covers the real 10 MB limit.
- #24: the sweep is scoped to cwd-in-checkout candidates, their descendants, and the per-checkout `METIS_SIDECAR_OWNER` environment tag, which is read from the environment and never from argv. With no METIS process running, nothing is signalled. The re-exec that drops an inherited tag is a real catch.

### Security: CLEAN
No injection surface added: the owner tag comes from `cksum` of the checkout path, spawn keeps `shell: false`, and the marker is spread last so a configured server env cannot spoof or remove it. `ps -E` and `/proc/*/environ` contents are only compared in-process and never logged.

### Tests: SUFFICIENT, and the fail-before claims were verified
I reverted only the four production files (agent-loop.ts, error-handler.ts, stdio-transport.ts, restart.sh) to origin/main. The new tests went red: 19 server tests and all 5 restart.sh process tests (with the kill shim, so nothing was signalled). After restoring the files, 69 of 69 server tests and 6 of 6 scripts tests passed. The BASH_ENV `kill` shim is a good isolation pattern.

### Gates
`gh pr checks 47`: every check passes. `e2e` and `windows` were skipped and nothing is pending. `api-outcome` and `postgres-adapter` passed. Locally, on the changed files: `shellcheck scripts/restart.sh` was clean, eslint was clean (the .mjs test is eslint-ignored by config), `tsc --noEmit` in server passed, and `pnpm changelog:verify` passed with three fragments. I did not re-run the full monorepo `pnpm test`.

### Documentation: UP TO DATE
Three `.changes/unreleased/` fragments, no CHANGELOG.md edit, no version bump.

### Findings (all non-blocking; see inline)
1. Medium: the cwd prefix match also covers nested `.claude/worktrees/*` checkouts (verified empirically).
2. Low: the self-exclusion covers ancestors, not the script's own subshells, when an ancestor is a root.
3. Low: on Windows, orphaned sidecars are no longer swept at all, and #48 does not record this.
4. Low (perf): the Linux environ scan is a pure-bash read loop over every process.

### Scope of this review
I read the full diff of all 13 files and the relevant surrounding code. I did not run `scripts/restart.sh` against the live machine: every script run used a temp-dir copy with a recording `kill` shim. I did not exercise the running app.

Issue #52, opened 2026-09-22
Generated-doc detail API returns raw exception text for failed generations


mgcronin:
**`GET /projects/:projectId/docs/:docId` returns the raw server exception text of a failed generation to the client.**

_Found while building the failed-document banner for #50._

## Where

- `server/src/routes/generated-docs.ts` — the `generateDocumentAsync` catch path stores `errorMessage: String(err)` on the row when a claimed generation throws.
- The detail route (`r.get("/:docId", …)`) returns `...doc`, so that string reaches the browser verbatim.

The socket path already avoids this: #254 sends only `genericFailureMessage("doc-generation")` over `job:failed` and keeps the detail in the server log. The REST detail payload bypasses that rule. `String(err)` can carry provider error bodies, file paths or SQL error text.

The UI does not render `errorMessage` for a `failed` document (the #50 banner deliberately shows generic text), so the exposure is in the API response only.

## Suggested fix

Either store a user-safe message in `errorMessage` and keep the detail in the log, or strip `errorMessage` from the detail response for `failed` documents (the `interrupted` flag added in #50 already carries the one case the UI distinguishes). Add a test asserting a thrown `Error("secret detail")` never appears in the `GET /:docId` body.

PR #53, opened 2026-09-22
fix: Resolve #50, #25 — fail interrupted doc generations; bound Phase-1 reasoning

### LINKED ISSUE #50: Documentation generation interrupted by a server restart stays "generating" forever
**A documentation generation interrupted by a server restart stays "generating" forever.** It is not resumed, retried or failed, so the document card shows the same progress indefinitely and the run's partial token spend is lost.

_Found during an end-to-end walkthrough on a local install: the dev server restarted (its file watcher reloaded after a `git pull`) 26 minutes into a full-project generation. More than 35 minutes later the document was still `status = generating` at the progress it had reached; no further model calls were made._

## Cause

Generation runs in-process, started from the request, not as a scheduler task: the `tasks` table holds rows for `publish-generated-document`, but none for generation itself. The scheduler's startup recovery (see `server/src/lib/docs-gen/generated-doc-outbox.ts`, "The scheduler's startup/periodic recovery owns pending rows if the process dies") therefore never sees it.

Any restart does this: a deploy, a crash, a container reschedule, a dev hot-reload.

## Acceptance criteria

- [ ] On startup, a generated document left in `generating` (or `pending`) by a previous process is detected
- [ ] It is either resumed, or marked failed with a clear "interrupted by a restart" message and a one-click regenerate — never left spinning
- [ ] A test simulates a process exit mid-generation and asserts the document does not stay `generating`
- [ ] Consider running generation as a scheduler task so the existing recovery covers it

mgcronin:
Closes #50
Refs #25. This covers the remaining work from the latest comment (bounding Phase-1 reasoning). #25 stays open because it still needs a live re-measurement of a complete run, and its "covers the RAG subsystem" criterion needs one too.

## #50: interrupted generations never stay `generating`

Generation runs in-process (`generateDocumentAsync`, fire-and-forget), so the scheduler's durable recovery never sees it.

- **Heartbeat.** While a run holds its claim it writes `generated_documents.updatedAt` every 60 s. The write is fenced on `status = generating` and the claim (`server/src/lib/docs-gen/interrupted-generations.ts`). It is stopped in a new `finally` in `generateDocumentAsync`.
- **Sweep.** `server.ts` runs it once at startup and then every minute. It does not depend on the scheduler, which can be disabled. It fails:
  - `generating` rows with no heartbeat for 5 minutes;
  - `pending` rows untouched for 15 minutes (a pending row cannot heartbeat because `updatedAt` is its claim fence).
  
  Each row is updated compare-and-set on the `updatedAt` value that was read. The sweep clears the dead run's claim, so a stalled run that wakes up later cannot overwrite the row. It also emits `job:failed`, so an open UI leaves the spinner. The trigger is a missing heartbeat, not "this process just booted", so another replica's live run is never failed.
- **One-click regenerate.** `POST /projects/:id/docs/:docId/regenerate` moves the row from `failed` to `pending` with a compare-and-set, so a double click starts only one run. It shares `/generate`'s 5/15-min rate limit and the router's `requireProjectAccess` + `project.update` checks. It then reruns the same document in place, and the Phase-1 fact cache means modules already extracted are not paid for again. The detail response gains `interrupted: boolean`. The UI shows a banner for failed documents ("interrupted" or a generic failure, never the raw error) with a **Regenerate** button.
- **Decision: I did not make generation a scheduler task.** An automatic resume after a restart would spend model tokens with no one watching. The scheduler can be disabled. The existing `regenerate-generated-document` handler only covers automatic runs (it needs `autoUpdate` plus a fingerprint fence). Failing the document and offering a one-click regenerate meets the acceptance criterion without those costs.

## #25: bound Phase-1 reasoning on thinking-by-default models

I checked DeepSeek's documentation for the Anthropic format (https://api-docs.deepseek.com/guides/thinking_mode, https://api-docs.deepseek.com/guides/anthropic_api). It toggles thinking with `thinking: {type: enabled|disabled}` and sets effort with `output_config: {effort: low|high|max}`. Thinking is on by default at effort `high`, and a request for `medium` is treated as `high`. `budget_tokens` is ignored.

- New tunable `DOCS_GEN_PHASE1_REASONING` (registry, db → env): `auto` (default) | `provider-default` | `off` | `low` | `medium` | `high`.
  - `auto` sends `reasoningEffort: "low"` only when the model served is a thinking-by-default one: `deepseek-v4-pro` / `deepseek-flash`, or a `claude-*` name on DeepSeek's endpoint. This reuses #41's allow-list.
  - For every other model it sends **nothing**, so Claude requests are unchanged.
- `anthropic-provider.ts`: adds `ChatOptions.disableThinking`, which sends `thinking: {type: "disabled"}` and no effort. On a DeepSeek `baseUrl`, an effort goes out with `thinking: {type: "enabled"}`, the documented value. Other endpoints keep `adaptive`, as before.
- The Phase-1 fact-cache key includes the reasoning setting only when it is not empty, so Claude cache keys do not change. `DOCS_GEN_REASONING_ALLOWANCE_TOKENS` (#41) still sets the cap, which stays as safety headroom.
- Only Phase 1 is bounded. Phase 2, claim extraction and the judge keep the provider default until a full run has been re-measured.

## Tests and the mutations that proved they can fail

Each mutation was applied, the named test file was run with `--retry=0`, and the change was reverted. **Every one went red.**

| Mutation | Result |
|---|---|
| DeepSeek toggle `"enabled"` → `"adaptive"` | 1 failed |
| drop the `disableThinking` branch | 1 failed |
| treat every endpoint as DeepSeek | 2 failed |
| resolver `auto` → `{}` for thinking models | 4 failed |
| resolver `auto` → effort for **every** model (Claude changed) | 3 failed |
| `off` → `{}` | 1 failed |
| unrecognised value falls back to `provider-default` | 3 failed |
| Phase-1 call site drops `...reasoning` | 2 failed |
| cache key drops `reasoning` | 1 failed |
| fixture key drops `disableThinking` | 1 failed |
| no heartbeat started after the claim | 1 failed (a live run gets swept) |
| heartbeat never stopped (`finally` removed) | 1 failed |
| sweep does not revoke the claim | 2 failed (the stale run overwrites the row) |
| sweep revokes the claim on `pending` rows too | 1 failed |
| no `pending` sweep / `pending` swept on the 5-min threshold | 1 failed each |
| no `job:failed` emit | 1 failed |
| no startup pass (interval only) | 1 failed |
| sweeper stop is a no-op | 1 failed |
| regenerate without the `status: failed` CAS | 1 failed |
| regenerate that never starts generation | 1 failed |
| `interrupted` flag true for any failure | 1 failed |
| UI: no banner / `interrupted` ignored / wrong URL / button never disabled / one headline for both cases | 1–2 failed each |

**One mutation survived**, and it is equivalent: removing `deletedAt: null` from the sweep's `findMany`. The per-row compare-and-set also filters `deletedAt: null`, so a deleted row is never updated either way.

The process-exit simulation is in `server/tests/generated-doc-interrupted-generation.test.ts`. It runs the real `generateDocumentAsync` against an in-memory `generated_documents` table. The synthesizer never returns, and the process "dies" by `vi.clearAllTimers()`, which removes its heartbeat. A new sweep then fails the row. The companion test advances time 15 minutes with a live heartbeat and asserts the row is **not** touched.

## Coverage (server, full suite, v8)

| File | Before (main) lines / branches | After |
|---|---|---|
| `routes/generated-docs.ts` | 98.45 / 91.19 | 98.16 / 91.37 |
| `ai/providers/anthropic-provider.ts` | 100 / 88.28 | 99.32 / 88.8 (full suite, before the last `isDeepSeekEndpoint` test was added) |
| `docs-gen/output-caps.ts` | 97.14 / 100 | 97.22 / 100 |
| `ai/fixtures/fixture-key.ts` | 100 / 100 | 100 / 100 |
| `docs-gen/interrupted-generations.ts` (new) | — | 100 / 88.88 |
| `docs-gen/docs-gen-reasoning.ts` (new) | — | 100 / 92.85 |

## Gate (run locally before the push)

`pnpm lint && pnpm typecheck && pnpm test` exited 0. Server: 1113 files / 17,377 tests passed. UI: 353 / 3,261. The private-terms scan covered the staged files. `ui` `pnpm lint` and `pnpm build`, `agents:verify`, `skills:verify` and `verify-changelog-fragment` all passed. There are two changelog fragments: `50-interrupted-doc-generation.md` and `25-docs-gen-bounded-reasoning.md`.

## Security

- **Access control:** the new endpoint is behind the router-level `requireProjectAccess` (a non-member gets a 404, which is tested) and `requirePermission("project.update")`, and it is rate-limited.
- **Scoping:** every query is scoped to `projectId` plus the id. There is no raw SQL or shell, and no dependency was added.
- **Evidence policy:** regenerate reuses the document's stored evidence policy, the same as automatic regeneration.
- **Filed #52:** `GET /docs/:docId` already returned the raw `String(err)` of a failed generation. This PR does not change that, and the new banner does not display it.

## Deferred

- A live re-measurement of a full `deepseek-v4-pro` run (tokens, time, truncation) against #25's table.
- Bounding reasoning for Phase 2, claims and the judge.
- A failed document whose automatic regeneration task is still `running` is only re-queued by the scheduler's existing 2-hour lease.


server/src/lib/docs-gen/interrupted-generations.ts
```diff
@@ -0,0 +1,139 @@
...
+  });
+  let failed = 0;
+  for (const doc of stale) {
+    const result = await prisma.generatedDocument.updateMany({
+      where: {
+        id: doc.id,
+        projectId: doc.projectId,
+        deletedAt: null,
+        status: doc.status,
+        updatedAt: doc.updatedAt,
```
mgcronin, line 104: Compare-and-set on the `updatedAt` that was read. If a live run heartbeats, or a pending row is claimed, between the `findMany` and this write, the update matches nothing and the row is left alone. That is what makes the sweep safe with more than one replica.

server/src/lib/ai/providers/anthropic-provider.ts
```diff
@@ -426,8 +429,15 @@ export class AnthropicProvider implements AIProvider {
...
-      body.thinking = { type: "adaptive" };
+    if (opts.disableThinking) {
+      // #25 — explicit OFF. Documented on the Anthropic API and on DeepSeek's
+      // Anthropic-compatible endpoint; an effort would contradict it, so none.
+      body.thinking = { type: "disabled" };
+    } else if (opts.reasoningEffort) {
+      // #25 — DeepSeek documents only `enabled`/`disabled` for the Anthropic
+      // format (https://api-docs.deepseek.com/guides/thinking_mode); Claude's
+      // `adaptive` is kept for every other endpoint.
+      body.thinking = { type: this.deepSeekEndpoint ? "enabled" : "adaptive" };
```
mgcronin, line 440: DeepSeek documents only `enabled`/`disabled` for Anthropic-format thinking (https://api-docs.deepseek.com/guides/thinking_mode). `adaptive` is kept for every other endpoint, so Claude requests are byte-identical to before.

mgcronin (commented):
Author notes on the parts a reviewer is most likely to question.

server/src/lib/docs-gen/holistic-synthesizer.ts
```diff
@@ -2018,6 +2022,9 @@ Extract ALL facts now. Be EXHAUSTIVE — every validation, every conditional, ev
         effectiveConfigHash,
         maxTokens,
         supportsCaching,
+        // Facts extracted at a different reasoning setting are different facts;
+        // omitted when empty so an unchanged (Claude) request keeps its key.
+        ...(Object.keys(reasoning).length > 0 ? { reasoning } : {}),
```
mgcronin, line 2027: Added to the key only when it is not empty, so existing Claude fact-cache entries keep their keys. DeepSeek deployments take one Phase-1 cache miss, because facts extracted at `low` effort are not the same as facts extracted at `high`.

server/src/lib/docs-gen/interrupted-generations.ts
mgcronin, line 104: **Low (tests), confidence high: this compare-and-set is untested.** I deleted `updatedAt: doc.updatedAt,` (M4) and all 15 tests in `generated-doc-interrupted-generation.test.ts` still pass. This guard is what keeps a sweep from failing a live run whose heartbeat lands between the `findMany` and this `updateMany` (on another replica, or after a slow write). The ARCHITECTURE.md entry names it as a guarantee.

Suggested test: have `findMany` return a stale snapshot, bump the stored row's `updatedAt` before `updateMany` runs, and assert the row is still `generating` and `failed === 0`.

server/src/lib/docs-gen/interrupted-generations.ts
```diff
@@ -0,0 +1,139 @@
...
+): () => void {
+  const timer = setInterval(() => {
+    prisma.generatedDocument
+      .updateMany({
+        where: {
+          id: docId,
+          projectId,
+          deletedAt: null,
+          status: "generating",
+          codeGraphHash: claim,
```
mgcronin, line 67: **Low (tests), confidence high: neither check in the heartbeat's filter is covered.** Removing `codeGraphHash: claim` (M5), or `status: "generating"` (M6), leaves all 15 tests green. The docstring says "a run that has been superseded or failed can never revive the row", but nothing enforces it. Today the sweep also clears the claim, so M6 is masked by M5's check, but that also means neither check is proven.

Suggested tests:
1. Start a heartbeat for `regenerating:old` on a row whose claim is `regenerating:new`, advance the timers, and assert `updatedAt` is unchanged.
2. Do the same on a `failed` row that still carries the heartbeat's claim.

server/src/routes/generated-docs.ts
```diff
@@ -505,6 +511,41 @@ export function generatedDocsRouter(): Router {
...
+          select: { status: true },
+        });
+        if (!existing) throw new AppError(404, "DOC_NOT_FOUND", "Generated document not found");
+        throw new AppError(
+          409,
+          "DOC_NOT_REGENERATABLE",
+          `Only a failed document can be regenerated (status: ${existing.status})`,
+        );
+      }
+      void generateDocumentAsync(docId, projectId).catch((err) => {
```
mgcronin, line 542: **Low/Medium (security, OWASP A01/A09), confidence medium, advisory.** Regenerate re-runs `generateDocumentAsync`, which calls `resolveEvidencePolicy(original)`. That runs the job as the **original creator's** stored principal, with their reference documents and ACL, not as `req.user`. `/generate` binds the policy to `req.user` via `createEvidencePolicy`.

The result is that user B (with `project.update`) can trigger a fresh generation grounded in user A's evidence set. The spend and the publication task are attributed to A, and nothing records that B asked for it. Automatic regeneration already runs as A, so this has precedent, but it was system-initiated and only for `autoUpdate` docs; this makes it user-initiated for every doc.

Options:
- (a) Rebind the policy to `req.user` on regenerate.
- (b) Allow it only for the principal or an admin.
- (c) At minimum, log or audit the requester alongside the doc id.

A related **Low (UX)** issue: if A has been deactivated or has lost `project.update`, `resolveEvidencePolicy` throws before the claim. The route has already returned 202, and the row goes back to a generic "Generation failed", so the Regenerate button loops. Pre-checking the policy in the route and returning a 403 or 409 that explains why would close this.

server/src/server.ts
```diff
@@ -418,6 +419,10 @@ export function createServer(opts: CreateServerOptions = {}): MetisServer {
       .recoverStaleRuns()
       .then(() => runner.dispatchQueued())
       .catch((err) => log.warn("Async runner recover failed", { error: (err as Error).message }));
+    // #50 — documentation generation runs in-process, so a restart orphans it.
+    // Fail any whose heartbeat stopped (now, then every minute) instead of
+    // leaving it `generating` forever.
+    startInterruptedGenerationSweeper();
```
mgcronin, line 425: **Low, confidence high.** Two issues here:
- **The wiring is untested.** Deleting this line survives the suite, because `createServer` skips the block under `NODE_ENV=test`. The sweeper's own startup pass is tested; this call site is not.
- **The stop handle is discarded,** so `shutdown()` never stops the interval. It is `unref`'d, so this does not block exit.

The sweep also runs on every replica rather than through the leader-elected singletons around line 88. The compare-and-set makes that safe, but it means N full-table scans a minute across the fleet. Consider registering it as a singleton job, which gives you a `stop()` handle and a leader-scoped lifecycle.

server/src/lib/docs-gen/interrupted-generations.ts
```diff
@@ -0,0 +1,139 @@
...
+        deletedAt: null,
+        status: doc.status,
+        updatedAt: doc.updatedAt,
+      },
+      data: {
+        status: "failed",
+        errorMessage: GENERATION_INTERRUPTED_MESSAGE,
+        // A `generating` row holds the dead run's claim here; clearing it means
+        // that run can never commit or fail over this row if it was only stalled.
+        ...(doc.status === "generating" ? { codeGraphHash: null } : {}),
```
mgcronin, line 111: nit (confidence high, no current impact): the in-process failure path restores `codeGraphHash: originalHash` (`generated-docs.ts`, catch block), but the sweep sets it to `null`, which drops the last good generation's fingerprint. Nothing else reads `codeGraphHash` today (I grepped server/src, ui and packages), so this changes no behaviour. The two paths do diverge, though, and a future reader of the hash would see `null` only on interrupted rows.

server/src/routes/generated-docs.ts
```diff
@@ -396,6 +400,8 @@ export function generatedDocsRouter(): Router {
     res.json({
       data: {
         ...doc,
+        // #50 — lets the UI explain a restart and offer a one-click regenerate.
+        interrupted: doc.status === "failed" && doc.errorMessage === GENERATION_INTERRUPTED_MESSAGE,
```
mgcronin, line 404: nit: `interrupted` is derived by exact string equality on the user-facing message. Rewording `GENERATION_INTERRUPTED_MESSAGE` would silently clear the flag on every row already stored, and the UI would fall back to the generic banner. Consider a stable marker, such as a code prefix or a `warnings` entry of kind `interrupted`, or at least a comment on the constant warning that its text is a persisted key.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking findings, 6 non-blocking

This is an automated analysis published as a COMMENT, not a human sign-off. The reviewing gh identity (mgcronin) is also the PR author, so GitHub cannot record an approval or a change request from it. The merge decision stays with a human.

### Requirements: PASS (#50), PARTIAL by design (#25 stays open)
- [x] #50: on startup, a document left `generating` or `pending` is detected. The sweep runs at boot and then every minute. It waits for 5 min without a heartbeat (`generating`) or 15 min untouched (`pending`), so "never spinning" holds within a bound of about 6 min rather than immediately.
- [x] #50: the row is marked failed with a clear "interrupted" message and a one-click Regenerate (`POST /docs/:docId/regenerate`, a compare-and-set from failed to pending).
- [x] #50: a test simulates a process exit mid-generation. It runs the real `generateDocumentAsync`, lets timers die, then runs the sweep.
- [x] #50: running generation as a scheduler task was considered and rejected, with a reason (unattended token spend, and the scheduler can be disabled).
- [x] #25: Phase-1 reasoning is bounded and configurable (`DOCS_GEN_PHASE1_REASONING`, default `auto`). Claude is unchanged by default. The larger cap is kept.
- [ ] #25: live re-measurement of a full run is deferred. The PR only references #25, which is correct.

**DeepSeek claims checked against the source.** I read the raw HTML of https://api-docs.deepseek.com/guides/thinking_mode. The toggle cell `{"thinking": {"type": "enabled/disabled"}}` is `colspan="2"`, covering both the OpenAI and Anthropic columns, and Anthropic-format effort is `output_config.effort` = `low/high/max`, with `medium` mapped to `high`. The anthropic_api page says `thinking` is supported and `budget_tokens` is ignored. The PR's choice of `enabled` over `adaptive` on DeepSeek's host is correct. I also confirmed that only `anthropic-provider.ts` honours `reasoningEffort`, and that the provider's base URL and `servedModelThinksByDefault` both read `ANTHROPIC_BASE_URL` from env, so the two do not disagree.

### Security: 1 advisory finding (Low/Medium)
- Regenerate runs under the stored principal of the original creator, not the user who asked for it. See the inline comment on `generated-docs.ts`.
- No injection surface. Project scope comes from `requireProjectAccess` plus `projectId` in every WHERE clause. The `interrupted` banner never renders raw `errorMessage` (#52 already tracks the existing leak).

### Tests: gaps found by mutation
I reverted the production code only and re-ran the tests, each file separately:
- `routes/generated-docs.ts` reverted to main: 6 tests fail (heartbeat, regenerate, interrupted flag).
- `holistic-synthesizer.ts` reverted to main: 3 tests fail.
- The provider's `enabled` toggle and `disableThinking` removed: 2 tests fail.

**Three mutations of my own survived** (15/15 tests pass), each on a concurrency guard that the PR description and ARCHITECTURE.md rely on:
- M4: the sweep's `updatedAt: doc.updatedAt` compare-and-set removed.
- M5: the heartbeat's `codeGraphHash: claim` check removed.
- M6: the heartbeat's `status: "generating"` check removed.

See the inline comments.

### Gate (as named in CLAUDE.md, scoped to the changed files)
- `eslint --max-warnings=0` on the 17 changed .ts/.tsx files passed.
- `tsc --noEmit` passed for server and ui.
- The 5 affected server test files passed (112 tests), and the UI banner test passed (5 tests).
- I did not re-run the full-monorepo `pnpm test`. CI ran it.
- CI: 14 checks pass, 2 are skipped (e2e, windows), 0 pending. `api-outcome` passes.

### Documentation: UP TO DATE
- Two `.changes/unreleased/` fragments; `CHANGELOG.md` is untouched, and no version was bumped.
- ARCHITECTURE.md covers the new module and config key; USER_GUIDE.md covers the regenerate flow.

### Scope of this review
I read the whole diff, `generateDocumentAsync` in full (claim, commit and failure fences), `resolveEvidencePolicy`, the provider factory and the fact-cache write path. The cache is upserted per module, so "reuses modules already analysed" holds. I did not exercise the running application.

Good work overall. The fencing design (a heartbeat keyed on the claim, a compare-and-set sweep, and revoking the dead run's claim so a stalled run cannot commit later) is sound, and the choice not to auto-resume is well argued.

PR #55, opened 2026-09-22
fix: Resolve #45, #54 — metis-server image starts on Postgres and writes its default data paths

### LINKED ISSUE #45: metis-server image cannot start against Postgres: Prisma client is generated for SQLite only

### LINKED ISSUE #54: metis-server image: default LanceDB path is not writable by the runtime user, and the smoke probe writes to /tmp instead
Found in review of #46 (finding 1), and re-measured on `origin/main` 9f310e2.

**The `metis-server` image's default LanceDB path is not writable by the user the server runs as.**

`server/src/lib/rag/vector-store.ts:1256` defaults the store to
`process.env.LANCEDB_PATH ?? path.resolve(process.cwd(), "data", "lancedb")`. `Dockerfile.server`
runs `CMD ["node", "server/dist/index.js"]` from `WORKDIR /app`, so the default is
`/app/data/lancedb`. `/app` is `root:root 755` and the server runs as uid 1001 (`metis`):

```
$ docker run --rm --entrypoint node metis-server:<9f310e2> --input-type=module \
    -e 'import fs from "node:fs"; fs.mkdirSync("/app/data/lancedb",{recursive:true})'
EACCES
```

The reviewer of #46 saw the LanceDB form of the same thing:
`Unable to created lance dataset at /app/data/lancedb: Permission denied (os error 13)`.

**Why CI did not catch it.** The `lancedb` probe in `scripts/lib/smoke-server-image.mjs` writes to
`/tmp/metis-smoke-lance`, which is writable, so the probe passes while the server's real default
path fails. The SQLite arm does the same thing (`DATABASE_URL=file:/tmp/metis-smoke.db`).

**The Helm chart has a related mismatch.** `deploy/helm/metis/values.yaml` mounts the uploads and
LanceDB PVCs at `/app/server/data/{uploads,lancedb}` and sets `readOnlyRootFilesystem: true`, but
because the process cwd is `/app`, the server's defaults are `/app/data/{uploads,lancedb}`. The PVCs
are never written to, and the first write to the default path hits a read-only filesystem.

## To do
- [ ] Make the server's default data paths writable by the runtime user, and put them where the
      chart mounts its volumes
- [ ] Point the smoke's storage probes at the server's real default paths, not at a scratch path
- [ ] Record the durable finding in `.claude/agent-memory/code-issue/`

mgcronin:
Closes #45
Closes #54

## What was broken (measured on `origin/main` 9f310e2, arm64 build)

1. **Postgres (#45).** With `DATABASE_URL=postgresql://…` against a pgvector container, the image exits at import with `PrismaClientInitializationError: The Driver Adapter @prisma/adapter-pg … is not compatible with the provider sqlite`. The image only ever generated the SQLite client.
2. **LanceDB default path (#54, finding 1 of #46's review).** The image runs from `/app`, which is `root:root`, so `<cwd>/data/lancedb` gives `EACCES` for uid 1001. Uploads (`<cwd>/data/uploads`) fail the same way. The smoke probe wrote to `/tmp`, so it passed anyway.
3. **Defect found while fixing #54: the SQLite default was split across two files.** With no `DATABASE_URL`, the migration guard (cwd `/app/server`) migrated `/app/server/dev.db`. The server's client (cwd `/app`) tried to open `/app/dev.db` and failed with `unable to open database file`. `/healthz` still returned 200, so every DB query failed on a server that reported healthy. This is the "write succeeds, read can't see it" shape, caused by the wrong storage location. #46's review saw `/readyz: database ok` only because it set `DATABASE_URL=file:/tmp/...`.
4. **Related, also fixed by the cwd change.** The Helm chart mounts its uploads and LanceDB PVCs at `/app/server/data/{uploads,lancedb}` with `readOnlyRootFilesystem: true`. The server's defaults resolved to `/app/data/...`, so the volumes were never written to.

## Design decision: one image, two generated clients

Prisma 7's `prisma-client-js` output is bound to one provider. It inlines that provider's query compiler (base64 wasm) in its `index.js` and checks `activeProvider` against the adapter in the constructor. I checked this in the generated code and in the image; the Prisma docs do not cover multiple clients. So:

- **`Dockerfile.server` (prod-deps stage).** Generate from `prisma/postgres/schema.prisma` and copy the output to `/app/server/prisma-clients/postgresql`. Then generate SQLite into the default `@prisma/client`. Both steps `grep` `activeProvider`, so a wrong schema fails the build. Files Node never loads are pruned from the copy (6.3 MiB left).
- **`server/src/lib/prisma.ts`.** New `resolvePrismaClientClass(provider)` picks the class using the same scheme rule that picks the adapter. `METIS_PRISMA_CLIENT_POSTGRESQL` (set by the image; must be an absolute path) points at the Postgres client. When it is unset, `@prisma/client` is used, so dev behaves as before. A path that is unloadable or has no `PrismaClient` fails loudly.
- **Checked in the image:** the relocated client resolves the same `@prisma/client/runtime`, so `Prisma.PrismaClientKnownRequestError`, `JsonNull`, `Decimal` and `sql` are the identical objects under both clients. `instanceof` checks elsewhere keep working.
- **Contradictory comment resolved.** The runtime stage now drops every query compiler in `@prisma/client/runtime`. Neither generated client loads them: `runtime/client.js` references no `query_compiler` file, and both smoke arms migrate and query without them. That saves about 29 MiB, more than the new client adds.
- **Runtime cwd.** Now `WORKDIR /app/server`, `CMD ["node", "dist/index.js"]`, and `/app/server/data` is created and owned by `metis`.

A per-provider build arg was rejected because `docker-compose.yml` and the smoke's default arm boot this same image on SQLite. Two images would double the `api` build.

## CI: Postgres arm (the smoke now runs two arms)

`scripts/lib/smoke-server-image.mjs` runs both arms by default. The `ci.yml` invocation is unchanged; only the comment was updated.
- **sqlite:** no `DATABASE_URL`, so the server's own default is what boots.
- **postgres:** creates a private docker network, starts `pgvector/pgvector:pg16` (the image the `postgres-*` jobs use), waits for `pg_isready -h 127.0.0.1` (the init restart is on the socket), then boots the image with `DATABASE_URL` pointing at it.
- **Probes in each arm** run in the container's WORKDIR and env, with no `--workdir`:
  - `database`: `prisma.user.count()` through `dist/lib/prisma.js`.
  - `migrations`: `prisma migrate status` using the image's CLI.
  - `lancedb`: `getVectorStore()` upsert, count, drop.
  - `uploads`: `resolveDocumentStorage()` write, read, remove.
  - Plus the existing mysql2 / better-sqlite3 / home / oracle-thick probes.
- Containers and the network are always removed, in order: server, then postgres, then network. Both arms run even if the first fails, so one run reports both.

### Proof it fails before the fix (real docker, local arm64)

| Image | sqlite arm | postgres arm |
|---|---|---|
| `origin/main` 9f310e2 | FAIL: `database` (unable to open database file), `lancedb` + `uploads` (EACCES mkdir `/app/data`), `mysql2`/`better-sqlite3` (cwd `/app`) | FAIL: exited before /healthz, `adapter-pg … not compatible with the provider sqlite` |
| this branch | PASS 8/8 | PASS 8/8 |
| this branch, WORKDIR/CMD reverted | FAIL database, lancedb, uploads, … | FAIL lancedb, uploads, … |
| this branch, `ENV METIS_PRISMA_CLIENT_POSTGRESQL` removed | PASS | FAIL: exited at import (same #45 error) |

## Mutations (each reverted; all went red)

- `prisma.ts`: ignore the override (`return PrismaClient`): 4 failed. Apply the override to every provider: 1 failed. Drop the "exports no PrismaClient" check: 1 failed.
- `smoke-server-image.mjs`:
  - Postgres URL not passed to the server: 1 failed.
  - Network not removed: 3 failed.
  - Arm failure ignored: 9 failed.
  - No `pg_isready` wait: 2 failed.
  - Probes pinned to `--workdir /app/server`: 2 failed.
  - LanceDB probe skips its read-back: 1 failed.
  - SQLite arm back on `file:/tmp/...`: 2 failed.
- `Dockerfile.server` reverted to `origin/main`: 6 new wiring tests failed. `smoke-server-image.mjs` reverted to `origin/main`: 16 of the new smoke tests failed.
- Image level: the two rows above, with WORKDIR reverted and with the env removed.

## Coverage (touched code)

| File | before (origin/main) | after |
|---|---|---|
| `server/src/lib/prisma.ts` | 100% lines / 92.3% branches | 100% / 94.4% |
| `scripts/lib/smoke-server-image.mjs` | 100% / 98% | 100% / 97.3% (the real docker/fetch adapters stay `c8 ignore`d, as before) |

## Image size

**CI amd64 (`api` run 35722462860): 1,045.4 MB, down from 1,067.7 MB at #39; the budget stays 1,170 MB.** Locally on arm64, `du -sm /app` went from 744 to 723 MiB: the Postgres client adds 6.3 MiB and removing the unused runtime compilers saves about 29 MiB. The figure is recorded in `docs/OPERATIONS.md` (table and #45 section) and in the `DEFAULT_MAX_SERVER_IMAGE_MB` comment in `verify-image-size.mjs`. `ci.yml` states no number. The same run's smoke step passed both arms, 8/8 probes each.

## Gate (local)

- `pnpm lint`: exit 0. The private-terms scan was re-run after staging, because new files are untracked until then.
- `pnpm typecheck`: exit 0. `scripts` `tsc --noEmit`: exit 0.
- `pnpm test`: exit 0. Server 1111 files passed, scripts 58, ui 352, shared 36.
- `pnpm agents:verify` and `pnpm changelog:verify`: pass.

## Security

- No dependencies added, so no audit delta.
- The smoke invokes docker with an argv array (no shell). The Postgres password is random per run, is never logged (a test asserts this), and the container is not published to the host.
- `METIS_PRISMA_CLIENT_POSTGRESQL` loads code from an operator-set path. Anyone who controls that env var already controls the process. The path must be absolute, and the directory in the image is root-owned, so the runtime user cannot write to it.

## Deferred / not done

- Under Helm's `readOnlyRootFilesystem`, cwd-relative dirs other than the two mounted volumes (`data/repo-extracts`, `data/repo-archives`) are still on the read-only root. That was already true before this PR; it needs a chart change (an emptyDir or a `/app/server/data` volume).
- With no `DATABASE_URL`, the SQLite default `/app/server/dev.db` is in the container layer, not under `data/`. `docs/OPERATIONS.md` now says to point `DATABASE_URL` into a volume. Changing `DEFAULT_SQLITE_URL` would change dev.
- `docker-compose.prod.yml` mounts no volume on the server, so LanceDB/uploads there are lost on container recreate. This PR doesn't change that.

Durable finding recorded in `.claude/agent-memory/code-issue/project_storage-probes-must-target-real-default-paths.md`: storage probes must target the server's real default paths, because a probe writing to a scratch path can pass while the real default fails.


server/src/lib/prisma.ts
```diff
@@ -106,10 +167,11 @@ export function redactDatabaseUrl(databaseUrl: string): string {
 export const boundDatabaseUrl: string = process.env.DATABASE_URL || DEFAULT_SQLITE_URL;
 
 const adapter = selectPrismaAdapter(boundDatabaseUrl);
+const PrismaClientForProvider = resolvePrismaClientClass(resolveDatabaseProvider(boundDatabaseUrl));
```
mgcronin, line 170: The client class uses the same scheme rule as the adapter on the line above, so the two cannot disagree. In dev (env unset) this is `@prisma/client`, exactly as before; the `postgres-adapter` CI job still covers dev-on-Postgres through `prisma generate`.

Dockerfile.server
```diff
@@ -133,11 +133,37 @@ COPY e2e/package.json ./e2e/
...
+ && grep -q '"activeProvider": "postgresql"' "${GENERATED}/index.js" \
+ && mkdir -p server/prisma-clients \
+ && cp -R "${GENERATED}" server/prisma-clients/postgresql \
+ && rm -f server/prisma-clients/postgresql/*.d.ts server/prisma-clients/postgresql/edge.js \
+          server/prisma-clients/postgresql/index-browser.js \
+          server/prisma-clients/postgresql/wasm-*-loader.mjs \
+          server/prisma-clients/postgresql/query_compiler_fast_bg.wasm \
+          server/prisma-clients/postgresql/schema.prisma \
  && pnpm --filter @metis/server prisma generate \
+ && grep -q '"activeProvider": "sqlite"' "${GENERATED}/index.js" \
```
mgcronin, line 166: Both `grep -q activeProvider` guards fail the build if a generate step ever reads the wrong schema. For example, if `prisma.config.ts` gets copied into this stage, a bare `generate` would follow `DATABASE_URL`.

Dockerfile.server
```diff
@@ -282,21 +309,17 @@ RUN set -eux; \
...
-            query_engine_bg.sqlite.* 2>/dev/null || true; \
-      # #34 — Prisma 7 renamed these to query_compiler_{fast,small}_bg.<db>.*, so
-      # the names above had stopped matching anything. The generated client
-      # (.prisma/client) carries its OWN copy of the compiler it uses; these are
-      # never loaded by it. Keep sqlite + postgresql (the two supported DBs).
-      rm -f query_compiler_fast_bg.cockroachdb.* query_compiler_fast_bg.mysql.* \
-            query_compiler_fast_bg.sqlserver.* query_compiler_small_bg.cockroachdb.* \
-            query_compiler_small_bg.mysql.* query_compiler_small_bg.sqlserver.* \
-            2>/dev/null || true; \
+      rm -f query_compiler_* query_engine_bg.* 2>/dev/null || true; \
```
mgcronin, line 322: This drops every query compiler from the runtime dir. `runtime/client.js` references no `query_compiler` file, and each generated client inlines its own base64 copy. Evidence: both smoke arms migrate and query without these files, locally and in CI.

Dockerfile.server
```diff
@@ -431,11 +458,24 @@ COPY --from=prod-deps  --chown=metis:metis /app/node_modules            ./node_m
...
+# (documents/storage.ts), archive extracts, and the SQLite default `file:./dev.db`.
+# Run from /app/server, not /app: /app is root-owned, so every one of those
+# defaults was EACCES for uid 1001, and the Helm chart mounts its uploads and
+# LanceDB volumes at /app/server/data/{uploads,lancedb} — where the server never
+# wrote. The migration guard already ran the Prisma CLI from /app/server, so the
+# SQLite default now names ONE file for the CLI and the client. `data/` is created
+# and owned by the runtime user so a fresh named volume mounted there inherits
+# that ownership.
+RUN mkdir -p /app/server/data && chown metis:metis /app/server/data
+WORKDIR /app/server
```
mgcronin, line 472: #54: this cwd change is the whole fix for LanceDB, uploads, the SQLite split-brain and the Helm mount mismatch. `/app/server/data` is chowned above so a fresh named volume inherits uid 1001.

scripts/lib/smoke-server-image.mjs
```diff
@@ -261,56 +420,106 @@ export async function runSmoke(opts, deps = {}) {
...
-        "/app/server",
-        name,
-        "node",
-        "--input-type=module",
-        "-e",
-        probe.code,
-      ]);
+      // No `--workdir`: the probe runs in the image's WORKDIR, the server's cwd, so
+      // every cwd-relative default resolves exactly as it does for the server.
+      const r = docker(["exec", name, "node", "--input-type=module", "-e", probe.code]);
```
mgcronin, line 438: There is deliberately no `--workdir`, so probes resolve cwd-relative defaults exactly as the server does. A test and mutation M8 pin this.

mgcronin (commented):
Author's self-review of 779760f. This is not an approval: the reviewing `gh` identity is the PR author.

**Checked**
- The Postgres arm fails on `origin/main` with the exact #45 error. The SQLite arm fails on `origin/main` on `database` / `lancedb` / `uploads`. Both arms pass on this branch, locally on arm64 and in CI on amd64 (run 35722462860, 8/8 probes each).
- Two image-level mutations each turn exactly the expected arm red: reverting WORKDIR/CMD, and removing `METIS_PRISMA_CLIENT_POSTGRESQL`.
- Size: 1,045.4 MB in CI against a 1,170 MB budget. The image is 22 MB smaller than at #39.
- `Prisma.*` identity across the two clients was checked in the image: error classes, `JsonNull`, `Decimal` and `sql` are identical.

**Found and fixed during self-review**
- A Dockerfile comment said the runtime compilers exist "for the newer `prisma-client` generator". I had not verified that, so I removed it.

**Left as is, noted in the PR body**
- Helm `readOnlyRootFilesystem`: cwd-relative dirs other than the two mounted volumes (`data/repo-extracts`, `data/repo-archives`) are still read-only. This was already the case before this PR; it needs a chart change.
- The SQLite default `dev.db` sits outside `data/`. OPERATIONS.md now tells operators to point `DATABASE_URL` into the volume.
- The Postgres arm's password is on docker's argv, same as the existing secrets (the nit from #46's review). It is random per run and the runner is ephemeral.

Blocking findings: 0.

scripts/lib/smoke-server-image.mjs
```diff
@@ -34,26 +51,78 @@ export const DEFAULT_TIMEOUT_S = 120;
...
+    code: [
+      'const { spawnSync } = await import("node:child_process");',
+      'const { createRequire } = await import("node:module");',
+      `const serverDir = "${SERVER_DIST}/..";`,
+      'const cli = createRequire(serverDir + "/package.json").resolve("prisma/build/index.js");',
+      'const r = spawnSync(process.execPath, [cli, "migrate", "status"], { cwd: serverDir, encoding: "utf-8" });',
+      'if (r.status !== 0) { process.stderr.write((r.stdout ?? "") + (r.stderr ?? "")); process.exit(r.status ?? 1); }',
+    ].join(" "),
+  },
+  {
```
mgcronin, line 98: **Medium, advisory (confidence: medium).** The Postgres arm probes LanceDB, but `deploy/helm/metis/values-prod.yaml:60` runs `VECTOR_STORE=pgvector`. That is the production configuration this bundle was meant to prove.

`rag/vector-store-pgvector.ts` runs raw SQL through `lib/prisma.ts`, so in the image it goes through the relocated Postgres client, and no probe exercises it. The pgvector image is already running in this arm. A `pgvector` probe (Postgres arm only) that sets `VECTOR_STORE=pgvector` via `docker exec --env`, upserts, counts and drops through `getVectorStore()` would close the gap. It fits the #54 lesson: probe the path production actually takes. Fine as a follow-up.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking, 5 non-blocking findings

This is an automated analysis published as a comment, not a human sign-off. The reviewing identity (`mgcronin`) is also the PR author, so GitHub cannot record an approval or a change request from it. A human still owns the merge decision.

## Code Review Summary — PR #55 (head 779760f)

### Requirements: PASS
**#45**
- [x] One image serves both providers. The image generates two clients, and `resolvePrismaClientClass` picks one by the same scheme rule as the adapter.
- [x] Contradictory query-compiler comment resolved. All runtime compilers are now dropped. I checked the generated `index.js`: it inlines its own compiler (`query_compiler_fast_bg.js` + base64 wasm), and its only package import is `@prisma/client/runtime/client.js`.
- [x] Postgres arm in the CI smoke (pgvector:pg16 on a private network, migrate status, a migrated-table read). CI run 35724383389 shows both arms `PASS ... 8 probe(s)`.
- [x] Proven to fail on main. I did not rebuild main's image. The author reports the exact #45 error, and #45's own body records it.

**#54**
- [x] Default data path writable. `WORKDIR /app/server`, and `data/` is owned by uid 1001.
- [x] Probe targets the real default. It goes through `getVectorStore()` / `resolveDocumentStorage()`, with no `--workdir`, no path argument and no `DATABASE_URL` on SQLite.
- [x] Durable finding recorded in `.claude/agent-memory/code-issue/`. `agents:verify` is green.

**Image size**
- [x] CI log shows `metis-server 1045.4 / 1170`. `docs/OPERATIONS.md` and the `verify-image-size.mjs` comment agree with it.

### Security: CLEAN
- `docker` is called with argv arrays and no shell.
- Secrets, including the Postgres password, are generated per run and never logged. A test asserts this.
- `METIS_PRISMA_CLIENT_POSTGRESQL` is operator-controlled env. It is required to be an absolute path, and the server fails loud if it cannot be loaded.
- The relocated client dir is root-owned in the image.
- No new dependencies.

### Design
Two clients sharing one runtime is sound. I confirmed the `Prisma.*` identity claim structurally: both generated clients take `PrismaClientKnownRequestError`, `DbNull`, `JsonNull` and the rest from `require('@prisma/client/runtime/client.js')`, which resolves to the same module from `/app/server`. The `/app/server` WORKDIR also fixes the two-SQLite-files defect and the Helm PVC paths. It is a good catch.

### Tests: SUFFICIENT
- Local: `prisma.test.ts` 27/27, and `smoke-server-image` + `image-build-wiring` 63/63.
- Reviewer mutation: I reverted the consumption site (`new PrismaClientForProvider` → `new PrismaClient`, prisma.ts:174).
  - No unit test fails.
  - ESLint `no-unused-vars` fails, and so would the Postgres smoke arm. That is acceptable (see inline).

### Gates (run by reviewer, in the PR worktree)
- `pnpm lint`: exit 0, including the private-terms scan.
- `pnpm typecheck`: exit 0.
- `pnpm agents:verify` and `pnpm changelog:verify`: pass.
- I did not re-run the full `pnpm test` locally. CI `api` ran it green on 779760f.
- `gh pr checks 55`: 14 pass, `e2e` and `windows` skipped, 0 pending.
- The branch is 1 commit behind `main` (086ffde, #53, docs-gen files only). `git merge-tree` reports no conflicts.

### Documentation: UP TO DATE
- The changelog fragment `.changes/unreleased/45-server-image-postgres.md` is present.
- The new OPERATIONS.md section is accurate.

### Findings (all non-blocking; details inline)
1. Medium: the Postgres arm does not exercise `VECTOR_STORE=pgvector`, the prod-values vector store, which runs raw SQL through the relocated client. Suggest a pgvector probe.
2. Medium: under the Helm default (no `DATABASE_URL` → SQLite) with `readOnlyRootFilesystem`, `/app/server/dev.db` cannot be created. This predates the PR. The deferred note says only "lost on recreate"; suggest a follow-up issue.
3. Low (informational): the prisma.ts:174 wiring has no unit test. Lint and the docker smoke catch a revert.
4. Low (informational): compose prod inherits `VECTOR_STORE=${VECTOR_STORE:-local}` from the base compose file, so compose-prod runs the JSON store, not LanceDB. This predates the PR.
5. nit: the OPERATIONS.md size table's ui/embeddings rows are stale against this run (206.7 / 588.8 MB). They predate the PR.

### Scope of this review
- Read all 14 changed files. I read `prisma.ts`, `Dockerfile.server` and `smoke-server-image.mjs` in full.
- Cross-checked the Helm templates and values, both compose files, `vector-store.ts`, `server-root.ts`, and the generated Prisma client.
- Grepped `server/src` for cwd-relative paths that the WORKDIR change could move.
- Did not build the image locally. Image behaviour rests on the CI `api` log for 779760f.

Dockerfile.server
mgcronin, line 472: **Medium, advisory; the problem predates this PR (confidence: medium — chart not rendered).** With no `DATABASE_URL`, the SQLite default is now `/app/server/dev.db`, on the root filesystem. The Helm chart's *default* values leave `DATABASE_URL` unset (`NOTES.txt:17`: "UNSET → embedded SQLite") and set `readOnlyRootFilesystem: true` (`_helpers.tpl:203`). So on a default `helm install`, the migration guard cannot create the file at all.

The PR's deferred list only says the file is "lost when the container is recreated", which understates it. It was equally broken before, when the guard wrote `/app/server/dev.db` too, so this is not a regression. Suggest filing it with the repo-extracts item: default the chart's SQLite URL into the data PVC, e.g. `file:/app/server/data/metis.db`.

server/src/lib/prisma.ts
```diff
@@ -106,10 +167,11 @@ export function redactDatabaseUrl(databaseUrl: string): string {
...
 
 export const prisma: PrismaClient =
   globalThis.__metisPrisma ??
-  new PrismaClient({
+  new PrismaClientForProvider({
```
mgcronin, line 174: **Low, informational.** Reviewer mutation: changing this to `new PrismaClient({` leaves every unit test green (prisma.test.ts 27/27). Only ESLint (`PrismaClientForProvider` is unused) and the docker smoke's Postgres arm would catch it. Both run in CI, so this is acceptable. Note it if this line is ever refactored so that the resolved class is still referenced but no longer constructed.

docker-compose.prod.yml
```diff
@@ -21,7 +21,8 @@ services:
     # Base file already defines a pg_isready healthcheck.
 
   server:
-    # Production uses the slim, postgres-only Dockerfile (issue #390).
+    # Production uses the slim Dockerfile.server (issue #390), which serves both
+    # Postgres and SQLite (#45).
```
mgcronin, line 25: **Low, informational; predates this PR.** This overlay merges `environment` with the base file, so it inherits `VECTOR_STORE: ${VECTOR_STORE:-local}` (`docker-compose.yml`). A compose-prod deployment therefore runs the JSON `LocalVectorStore`, not LanceDB, unless the operator overrides it. With `volumes: !override []` it also persists nothing (already on the deferred list). Worth folding into the same follow-up.

docs/OPERATIONS.md
```diff
@@ -323,7 +323,7 @@ The release pipeline ships three images:
 | Image | Built from | Current size | Budget |
 |---|---|---|---|
 | `metis-ui` | `Dockerfile.ui` (Next.js 15 standalone, alpine) | **~197 MB** | ≤ 350 MB ✅ |
-| `metis-server` | `Dockerfile.server` (`node:22-trixie-slim`, glibc, prod-only deps, slimmed) | **~1,068 MB** (amd64, measured #39) | ≤ 1,170 MB — its own budget, see below |
+| `metis-server` | `Dockerfile.server` (`node:22-trixie-slim`, glibc, prod-only deps, slimmed) | **~1,045 MB** (amd64, measured #45) | ≤ 1,170 MB — its own budget, see below |
```
mgcronin, line 326: nit, predates this PR: the neighbouring rows are stale against the same CI run (35724383389): `metis-ui` 206.7 MB (table says ~197) and `metis-embeddings` 588.8 MB (table says ~423). This row is correct. Optional.

PR #64, opened 2026-09-22
fix: Resolve #17, #16 — call graph binds on evidence; Deep Ingest keeps /healthz responsive

### LINKED ISSUE #16: Deep Ingest blocks the event loop — /healthz stops responding for minutes
**While a Deep Ingest runs, the whole API is unresponsive.** `/healthz` timed out (10s) for about 225 seconds during ingest of this repository (~3,400 files, 24k symbols). The process sat near 80% CPU; no request logs were written until it finished.

## Impact

- Any health check or load balancer marks the server down during every ingest.
- Every other user's requests — chat, page loads — stall for the duration.
- Server logs go silent, so it is indistinguishable from a hang.

## Likely cause

Parsing and graph construction run synchronously on the main thread. The later symbol-embedding pass (22k symbols, ~20 minutes) did *not* block — `/healthz` answered in under 50ms during it — so it is the parse/graph phase specifically.

## Acceptance criteria

- [ ] `/healthz` stays responsive (e.g. p99 < 1s) throughout a Deep Ingest of this repository
- [ ] CPU-bound ingest work runs off the main thread (worker threads) or yields regularly
- [ ] A test or benchmark that fails if a representative ingest blocks the loop beyond a threshold

### LINKED ISSUE #17: Code graph resolves calls by bare name — Array.prototype methods and test helpers dominate "most referenced"
**The call graph attributes every `.join()` in the codebase to whatever local function happens to be named `join`.** The Code Overview's "Top Symbols by In-Degree" for this repository:

| Rank | Symbol | In-degree |
|---|---|---|
| 1 | `ui/src/components/analysis/ClarificationDialog.tsx::join` | 2,853 |
| 2 | `ui/tests/publish-page-live-confirm.test.tsx::mock` | 1,907 |
| 3 | `ui/src/app/(authed)/projects/[id]/rule-sets/page.tsx::trim` | 1,606 |
| 4 | `server/src/lib/mcp/mcp-service.ts::MCPRegistryService::test` | 1,246 |
| 6 | `ui/tests/requirement-history-tab.test.tsx::beforeEach` | 1,030 |
| 8 | `.../settings/api-keys/page.tsx::report::replace` | 834 |

Genuinely central symbols (`AppError`, `requirePermission`, `apiFetch`) rank below them. The generated summary also quotes the doc comment of the mis-ranked `test` method as if it described the project.

## Impact

Everything built on in-degree inherits the error: god-node detection, entry points, impact-analysis caller sets, and documentation generation.

## Acceptance criteria

- [ ] Calls on a receiver (`x.join()`) are not resolved to an unrelated same-named function without evidence the receiver is that type/module
- [ ] Built-in/global methods and test-framework globals (`describe`, `it`, `beforeEach`, `vi.mock`) are never resolved to local symbols by name alone
- [ ] Test files are excluded from, or reported separately in, the overview's top-symbol ranking
- [ ] A fixture test: two files each defining `join`, callers using `array.join()` — neither gains in-degree

mgcronin:
Closes #17
Closes #16

Both issues are about the code-graph ingest, so they ship together in one PR. #17 is about which calls resolve to which symbols. #16 is about the ingest holding the event loop while it persists them.

## Measured on this repository

Corpus: `git archive` of `9f310e2` (3,384 files parsed, 22,181 symbols, 385,163 edges). The run used a scratch SQLite DB and `server/scripts/bench-ingest.ts`, which is new in this PR. The benchmark probes `/healthz` from a **worker thread** every 250 ms, with a 10 s timeout, the way a load balancer would. It also records main-thread stalls, including the gap still open when the ingest returns. The first version of the harness probed from inside the same process. That prober was starved along with everything else, so it reported the unfixed code as healthy (p99 17 ms). A probe on the loop under test can't see that loop's stalls.

### #16: event loop

| | before (origin/main) | after |
|---|---:|---:|
| ingest wall time | 458 s | 81 s |
| longest main-thread stall | **452.6 s** (one stall, 6 s → end) | 212 ms |
| `/healthz` probes answered | 24 | 246 |
| `/healthz` timeouts (10 s) | **44** | 0 |
| `/healthz` p50 / p99 / max | — (only the parse phase answered) | 73 / 177 / 248 ms |

**Cause.** Parsing turned out not to be the problem. The parse loop awaits `fs.readFile` for each file, and that await lets the event loop turn. The cause is the SQLite adapter. `@prisma/adapter-better-sqlite3` is synchronous, so every awaited Prisma call settles on the microtask queue. A loop of `await prisma.x.create()` never gives the event loop a turn, and the persist phase was about 420k of those calls with one commit each.

**Fix:**
- `event-loop-yield.ts`: a time-sliced `maybeYield()` that yields with `setImmediate` once 50 ms have passed. It is called for each parsed file, symbol, edge and rationale finding.
- Each file's symbol deletes and inserts now run in a single batch `$transaction`, and edges go in through `createMany` in batches of 500. This cuts the number of synchronous commits by about 40x. With yielding alone, individual fsync-bound commits still stalled for 4–7 s (p99 4.4 s).
- `SchemaGraphWriter` now yields before each row it writes. The ORM pass over this repo's `schema.prisma` writes about 2.2k symbols and 3.7k edges one row at a time. That pass on its own held the loop for about 8 s.

### #17: top symbols by in-degree

| # | before | in° | after | in° |
|--:|---|--:|---|--:|
| 1 | `ClarificationDialog.tsx::join` | 2922 | `error-handler.ts::AppError` | 874 |
| 2 | `publish-page-live-confirm.test.tsx::mock` | 1915 | `api-client.ts::apiFetch` | 457 |
| 3 | `rule-sets/page.tsx::trim` | 1616 | `require-permission.ts::requirePermission` | 424 |
| 4 | `mcp-service.ts::MCPRegistryService::test` | 1258 | `ai/types.ts::AIProvider` | 402 |
| 5 | `mcp-docker-stdio.test.ts::waitFor` | 1077 | `audit-service.ts::audit` | 295 |
| 6 | `requirement-history-tab.test.tsx::beforeEach` | 1037 | `logger.ts::createChildLogger` | 284 |
| 7 | `error-handler.ts::AppError` | 874 | `ai/types.ts::ChatMessage` | 248 |
| 8 | `api-keys/page.tsx::report::replace` | 835 | `config-service.ts::__resetConfigSingleton` | 173 |
| 9 | `types.d.ts::Request` | 521 | `connectors/types.ts::ConnectorError` | 161 |
| 10 | `api-client.ts::apiFetch` | 457 | `ai/types.ts::ChatResponse` | 155 |
| 13 | `admin/auth.ts::adminAuthRouter::p` | 394 | `schema-graph.ts::SchemaGraphWriter` | 125 |
| 14 | `service-config-store.test.ts::makePrisma::findFirst` | 373 | `ai/config.ts::loadAIConfig` | 117 |

The in-degrees of genuinely central symbols did not change: `AppError` 874 → 874, `apiFetch` 457 → 457, `requirePermission` 424 → 424, `audit` 295 → 295, `createChildLogger` 285 → 284. Resolved `calls` edges went from 76,519 to 60,723 (−15,796). About 12.7k of those came from the mis-bound names listed above alone.

**Rules** (`server/src/lib/code-graph/call-resolution.ts`; the parsers now record each member call's receiver, and TS import statements record the names they bind):
- **Bare calls** `foo()`: a name the file imports from the stdlib or a test framework (`import { join } from "node:path"`, `vitest`, `@testing-library/*`) never binds, and that is checked first. Otherwise the call binds to a non-method definition in the same file (a same-file method too in Java/C#, where a bare call is an implicit `this.`). Next comes a unique **top-level, non-method** match in an imported file. After that, the project-wide fallback applies only to **top-level, non-method** declarations in the **same language family**, and never to that language's runtime or test-framework globals (`describe`, `beforeEach`, `setTimeout`, `Error`, …; Python and Go builtins only in Python and Go).
- **Member calls** `x.foo()` never use the project-wide fallback. They can bind to:
  - `this.`/`self.` → a method in the same file, then in imported files. `super.`/`base.` skip the overriding method.
  - `Class.m()` → a member of that class.
  - `mod.f()` → a top-level function of that module (matched by file stem, `index` → its directory, Go → package directory).
  - any other receiver → a method in this file or an imported one. Two exceptions:
    - Built-in prototype names (`join`, `trim`, `test`, `replace`, …) never bind.
    - Collection-style names that are also common on domain types (`get`, `update`, `insert`, `delete`, `count`, …) bind only when the receiver is named after a class in an imported file (`orderMapper.insert()` → `OrderMapper::insert`).
  - A receiver that is a runtime namespace **of the caller's language** (`Math`, `JSON`, `vi`, `console`; Go `errors`; Python `os`, …) never binds, unless the file imports a project module of that name. Runtime-imported names never bind.
- New import evidence: `@/…` / `~/…` aliases resolve as `<ancestor>/src/…` (Next.js `@/*`). Java single-type imports resolve to `…/pkg/Class.java`. Without these two, the stricter member rule cost recall on the UI and on Java.
- **Overview:** symbols in test files are left out of the top-20 (pages through the ranking), and a note says how many were skipped. The summary's rationale therefore no longer quotes a test helper's doc comment.

### Things that depend on in-degree
- **God-nodes / overview summary:** the table above. The rationale in the summary now comes from product symbols.
- **Entry points** (zero inbound `calls` + path pattern): still 10 found on this repo (the list is capped at 10). `overview.entry-points.test.ts` passes.
- **Impact-analysis caller sets / blast radius, docs-gen (`code-graph-summary`, `discovery-agent`, `holistic-synthesizer`), `query-service`, `search-code-graph`:** these read `toSymbolId` and needed no code change. Their tests pass in the full suite. They now traverse fewer false edges: `.join()` is no longer a caller of `join`. The cost is losing edges that a globally unique name used to guess right with no evidence behind it.

## Tests — each seen to fail

Against unmodified `origin/main` source, the new tests fail. The issue's fixture (two files defining `join`, callers using `array.join()`), unique-method-on-unknown-receiver, globals/runtime-imports, Java-import, the event-loop test (a 4.6 s stall against the 500 ms limit), the overview test-file exclusion and the SchemaGraphWriter yield test are all red.

Mutations. For each one: apply, run the five code-graph test files with `--retry=0`, confirm red, restore.

| # | mutation | went red |
|--:|---|---|
| M1 | ignore the receiver (member calls resolved like bare) | 7 tests incl. issue fixture |
| M2 | drop the built-in method-name guard | issue fixture + unit |
| M3 | drop the global-name guard | globals fixture + unit |
| M4 | drop the runtime-import guard | globals fixture + unit |
| M5 | allow fallback to `method` symbols | top-level-only unit |
| M6 | allow fallback to nested declarations | top-level-only unit |
| M7 | allow cross-language fallback | language-family unit |
| M8 | drop the global/runtime receiver guard | receiver unit |
| M9 | no `Class.`/`module.` evidence | 4 incl. Go package fixture |
| M10 | `super.` binds the overriding method | super/base unit |
| M11 | Go functions not member-capable | Go unit |
| M12 | `node:` specifiers not treated as runtime | fixture + unit |
| M13 | Java imports unresolved | Java fixture |
| M14 | `@/` aliases unresolved | alias fixture |
| M15 | yielder never yields | event-loop ingest test (3.7 s) + 2 units |
| M16 | budget boundary off by one (`<=`) | yielder unit |
| M17 | slice never restarts | yielder unit |
| M18 | SchemaGraphWriter edge write without yield | writer unit |
| M19 | overview ranks test files | overview test |
| M20 | TS parser drops receivers | parser + 2 fixtures |
| M21 | TS named imports not captured | parser + globals fixture |

The first event-loop test used one file with 1,500 calls. Once the batching was in, it **stayed green with the yielder removed**, because batching alone hid the cost. It now uses 300 files, where the total synchronous work comes to more than 1 s however it is batched. M15 shows that version goes red.

## Coverage (touched files, lines; code-graph + ingest-wiring suites)

| file | before | after |
|---|---:|---:|
| `call-resolution.ts` (new) | — | 100% |
| `event-loop-yield.ts` (new) | — | 100% |
| `ingest.ts` | 97.76% | 97.43% |
| `overview.ts` | 95.27% | 95.83% |
| `parsers-tree-sitter.ts` | 85.76% | 86.40% |
| `parsers.ts` | 98.56% | 98.56% |
| `schema-graph.ts` | 100% | 100% |

## Gate (run locally, worktree off `e903a00`)
- `pnpm lint`: exit 0.
- `pnpm typecheck`: exit 0.
- `pnpm test`:
  - server: 1115 files passed, 17,451 tests.
  - ui: 353 files, 3,261 tests.
  - shared: 698. ui-kit: 7. copilot-svc: 78. embeddings-svc: 147.
  - scripts: 58 files, 2,343 tests. In the fan-out run, `scripts/lib/gate-cwd-anchoring.test.mjs` timed out once (5 s limit, scanning the tree twice under machine load from concurrent agents). It passed on an isolated re-run, and the whole scripts suite passed on a re-run. This PR does not touch `scripts/`.
- `node scripts/verify-changelog-fragment.mjs --base origin/main`: passes. There are two fragments, one per issue.

## Security
No new dependencies. No SQL is built from input; every write goes through Prisma. Alias and Java import resolution only look up paths in the in-memory set of parsed files and never touch the filesystem. `bench-ingest.ts` refuses any `DATABASE_URL` that isn't `file:`, and it evaluates a constant prober source in its worker.

## Limits and deferrals
- Member calls on an **unknown receiver whose type isn't imported** no longer bind. Examples: DI-injected services in Java that live in the same package without an import, or Go methods across packages. Before, these bound only when the method name was unique project-wide. There is no type inference here. The loss is deliberate and measured (−15.8k resolved edges in total, most of them false).
- Namespace imports whose alias differs from the module stem (`import * as scannerApi from "@/lib/scanner-api"`) resolve only through the member-in-imported-file rule, not through the stem rule.
- CommonJS `require` destructuring, Python `from os.path import join`, Go stdlib import paths and Java static imports are not treated as runtime-import evidence. The language-family rule and the top-level rule still cover most of the Python cases. This is tracked in #65.
- The legacy regex parsers, used only when tree-sitter fails to initialise, record no receivers.
- The rationale pass (`finding.findFirst` with `evidence contains`) is still one full scan per finding. Each scan is under 300 ms and it yields between findings, so it isn't a stall, but it is slow.
- SQLite journal mode is untouched (rollback journal, `synchronous=FULL`). WAL would cut commit cost further, but it is a database-wide change and belongs in its own issue.

## Review follow-up (`6bcfa16`)

This commit addresses review [5278768924](https://github.com/openzigs/metis/pull/64#pullrequestreview-5278768924). It fixes the blocking finding and takes findings 2–6. Finding 7 is filed as #65.

**Measured on the same corpus** (`git archive f9e65e3`: 3,394 files, 22,265 symbols, 387,099 edges; scratch SQLite; `bench-ingest.ts`):

| | `f9e65e3` (review) | `6bcfa16` |
|---|---:|---:|
| resolved `calls` edges | 61,154 | 60,983 |
| `/healthz` p50 / p99 / max | — | 71 / 141 / 174 ms, 301 probes, 0 timeouts |
| longest main-thread stall | — | 116 ms (none over 200 ms) |

The edge-level diff between the two: **193 edges lost, 22 gained.**
- **Lost:** every one is a bare call bound to a `method` symbol, which TS can't do. For example:
  - `getKeyDef(key)` inside `config-service.ts` bound to `ConfigService::getKeyDef` instead of the imported function.
  - Local `search(...)` / `upsert(...)` variables bound to `HybridCodeSearch::search` / `LocalVectorStore::upsert`.
  - `useQuery(...)` from `@tanstack/react-query` and `useState` setters bound to spurious top-level "method" symbols that the TS walker records at call sites (a separate parser quirk, not changed here).
- **Gained:** those same calls now land on the right target: `key-registry.ts::getKeyDef` +11, and `audit-service.ts::audit` +8, where an object-literal method `audit` in `prisma-adapter.ts` had shadowed the import.

Top-symbol table after (overview, test files excluded): `AppError` 876, `apiFetch` 458, `requirePermission` 425, `AIProvider` 402, `audit` 303, `createChildLogger` 285, `ChatMessage` 248, `__resetConfigSingleton` 173, `ConnectorError` 161, `ChatResponse` 155. God-nodes 20, entry points 10 (capped), both unchanged.

**Java/MyBatis recall (finding 3).** Corpus: `mybatis/jpetstore-6` @ `1ffdf1e`, 3,197 `calls` edges.

| `e903a00` (main) | `f9e65e3` | `6bcfa16` |
|---:|---:|---:|
| 792 | 828 | 828 |

jpetstore's mapper methods are named `getAccount`, `insertOrder` and so on, so none hit the built-in list, and this corpus can't show the loss the reviewer probed. That shape is now covered by a fixture: `orderMapper.insert()` / `orderMapper.update()` on an imported `OrderMapper` bind, and `cache.put()` on a `java.util.Map` field does not. A first attempt that let *any* imported file's method count as evidence was measured and rejected. On this repo it bound 398 `get` calls, mostly false: `analyses.get(id)` on a local `Map` became a caller of `ConfigService::get` (153 edges) only because the file imports config-service.

**Mutations for this commit.** Each was applied alone and run with `--retry=0` against `call-resolution.test.ts` + `ingest.test.ts`, and every one went red:

| # | mutation | red |
|--:|---|--:|
| R1 | drop the runtime-import short-circuit in `resolveBare` | 4 |
| R2 | same-file step reaches methods in every language | 2 |
| R3 | imported-file step reaches any kind | 3 |
| R4 | imported-file step reaches nested functions | 1 |
| R5 | no implicit-`this` languages (Java/C#) | 1 |
| R6 | bare index keeps methods (a method declared first hides the function) | 2 |
| R7 | collection names use the ordinary same-file/imported rule | 2 |
| R8 | collection names treated as strict built-ins (the reviewed behaviour) | 4 |
| R9 | global call names not scoped by language | 1 |
| R10 | global receivers not scoped by language | 1 |
| R11 | no import override for a runtime receiver name | 1 |
| R12 | runtime-imported receiver not checked first | 2 |
| R13 | receiver name not compared with the class name | 1 |
| R14 | receiver-named class need not be imported | 1 |

**Event-loop yields (finding 2).** There is now one test per yielding phase, with the synthetic cost charged only to that phase's writes. I removed each `await maybeYield()` in turn:
- pass 1 (symbols): red, 1,889 ms gap
- pass 2 (edges): red, 1,817 ms gap
- rationale: red, 1,219 ms gap
- parse loop: green. This is by construction: every iteration already awaits `fs.readFile`, which is a real I/O turn. The yield there is belt-and-braces, and no test can tell it apart.

**Gate, run in the worktree at `6bcfa16`'s tree:**
- `pnpm lint` exit 0.
- `pnpm typecheck` exit 0.
- `pnpm test` exit 0:
  - server: 1,115 files, 17,476 tests.
  - ui: 353 / 3,261. shared: 698. scripts: 2,343. copilot-svc: 78. embeddings-svc: 147. ui-kit: 7.
- `verify-changelog-fragment --base origin/main` passes.
- Coverage (lines): `call-resolution.ts` 100%, `ingest.ts` 98.13%.

**`bench-ingest.ts` (finding 6).** It now refuses a database that already holds any project, which the live dev database always does. It deletes its rows in a `finally`. I checked the refusal against a scratch DB seeded with one project.


server/src/lib/code-graph/ingest.ts
```diff
@@ -643,26 +666,45 @@ async function persistParsed(
...
-    // Wipe previous symbols+edges for this file so we don't accumulate stale rows.
-    await prisma.codeEdge.deleteMany({ where: { codeGraphId, filePath: file.filePath } });
-    await prisma.codeSymbol.deleteMany({ where: { codeGraphId, filePath: file.filePath } });
+    await maybeYield();
+    // Wipe previous symbols+edges for this file so we don't accumulate stale rows,
+    // and insert its symbols — one batch transaction per FILE (#16). On SQLite a
+    // statement outside a transaction is its own commit, and each commit is a
+    // synchronous journal write + fsync on the event-loop thread; per-row commits
+    // made the persist phase ~420k of them for this repository.
+    const results = await prisma.$transaction([
```
mgcronin, line 686: Why a batch transaction and not just yielding: with the yielder alone, the full-repo bench still showed single fsync-bound commits stalling 4–7 s (p99 4.4 s), because on better-sqlite3 each statement outside a transaction is a synchronous commit on the event-loop thread. One commit per file (plus createMany for edges) cut the commit count ~40x and wall time 458 s → 81 s. The batch form (not an interactive transaction) means no yield happens while the write lock is held.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,658 @@
...
+ *                 identifier (`this`, `arr`, `Foo`) for a member call, or
+ *                 {@link COMPLEX_RECEIVER} for any other receiver expression.
+ */
+export function resolveEdgeTarget(
+  name: string,
+  receiver: string | undefined,
+  site: ResolutionSite,
+  index: ResolutionIndex,
+): string | null {
+  if (receiver === undefined) return resolveBare(name, site, index);
```
mgcronin, line 664: The core of #17: only BARE calls ever reach the project-wide unique-name fallback. Every member call below is resolved from receiver evidence (this/self, Class., module stem / Go package dir, or a method in this file or an imported one). M1 in the PR body (route member calls through resolveBare) turns 7 tests red, including the issue's fixture.

mgcronin:
CI note: the first push failed **Semgrep** (`javascript.lang.security.audit.unsafe-formatstring`) on a template-literal `console.log` in `server/scripts/bench-ingest.ts`. f9e65e3 switches it to a constant format string. On a plain re-read at f9e65e3, all checks pass and none are pending (api 26m46s, api-outcome pass, postgres-adapter, ui, Semgrep, CodeQL). e2e and windows are skipped by workflow rules.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,658 @@
...
+  return uniqueAcross(site.importedFiles, index.fileToMemberIndex, name)?.id ?? null;
+}
+
+function resolveBare(name: string, site: ResolutionSite, index: ResolutionIndex): string | null {
+  // 1. Same-file definition wins outright — lexical shadowing.
+  const local = index.fileToNameIndex.get(site.filePath)?.get(name);
+  if (local) return local.id;
+
+  // 2. Unique match across the files this file imports.
+  if (site.importedFiles.length) {
```
mgcronin: **BLOCKING — Medium, high confidence. Bare calls to runtime- or test-imported names still bind to project METHODS by name.**

`resolveBare` checks `runtimeImports`/`GLOBAL_CALL_NAMES` only at step 3 (L643). Steps 1 and 2 run first, and both look names up in `fileToNameIndex`, which `indexSymbol` fills with **every** kind, methods and nested functions included. I probed it directly: with `runtimeImports = {join}` and importedFiles including a file whose only `join` is `PathUtils::join` (kind `method`), `resolveEdgeTarget("join", undefined, site, idx)` returns the method id. A bare `beforeEach()` in a file that imports a helper module defining `beforeEach` binds the same way.

The same happens inside one file. A `path-utils.ts` that does `import { join } from "node:path"` and calls bare `join(a, b)` from its own `PathUtils.join` method gets a self-edge to the method at step 1.

This contradicts #17 AC2 and the PR's own claim to have fixed `import { join } from "path"` binding to a project `join`. The fixture test in ingest.test.ts misses it because `thing.test.ts` neither imports `helpers.ts` nor defines `join` itself.

Suggested fix, small:
1. Short-circuit `site.runtimeImports?.has(name)` at the top of `resolveBare`. A name the file explicitly imports from `node:*` or vitest cannot be another project file's symbol.
2. In steps 1 and 2, skip `kind === "method"` and non-top-level symbols, since a bare call cannot reach a method. That is the same rule L644–652 already applies to step 3.
3. Add a fixture: the caller imports a project file that defines a `join` method, and the caller also imports `join` from `node:path`.

mgcronin (commented):
Recommendation: do not merge — 1 blocking finding (small, well-scoped fix), plus 2 medium and 4 low non-blocking findings.

This is an automated analysis published as a COMMENT, not a human sign-off. The reviewing gh identity is also the PR author, so GitHub cannot record an approval or change request from it. A human still owns the merge decision.

## Code Review Summary — PR #64 (#17, #16)

### Requirements: PARTIAL
- [x] #17 AC1: receiver calls (`x.join()`) no longer bind to unrelated same-named symbols. The issue's fixture is present and passes.
- [ ] #17 AC2: built-ins and names imported from the runtime or a test framework can still bind to project symbols through the same-file and imported-file steps, which run **before** the runtime-import guard. Reproduced below. **(blocking)**
- [x] #17 AC3: test-file symbols are left out of the overview's top-20, with a skipped-count note.
- [x] #17 AC4: fixture test (two `join`s, `array.join()` callers) is present and fails on base.
- [x] #16 AC1/AC2: ingest yields every 50 ms and writes in batches. The bench measurements (worker-thread prober: p99 177 ms, 0 timeouts) are plausible and the method is sound. I did not re-run the bench.
- [~] #16 AC3: the loop-blocking test exists and fails on base, but it guards only 1 of the 4 yield sites (see inline).

### Verification I ran (fresh detached worktree at f9e65e3, since removed)
- I reverted the production files only (ingest/overview/parsers/parsers-tree-sitter/schema-graph back to e903a00). **12 of the new tests went red**, including the #17 fixture, the unique-method, globals, Java-import, event-loop (12 s timeout) and overview-exclusion tests. The author's fail-before claim holds.
- Mutation check of the #16 test, removing one `await maybeYield()` at a time in ingest.ts. Pass-1 (L680) goes **red**. The parse loop (L368), the edge loop (L781) and rationale (L1929) all stay **green**.
- Direct probe of `resolveEdgeTarget`: a bare `join` imported from `node:path` resolves to `PathUtils::join` (a **method**) in an imported project file. A bare `beforeEach` resolves to an imported helper. `repo.update()` on an imported `Repo` resolves to null.
- `pnpm --filter ./server typecheck`: exit 0. `eslint` on all changed server files: clean. `vitest run` on tests/lib/code-graph, tests/ingest-*, impact-analysis and docs-gen: 27 files, 381 passed.
- CI on f9e65e3: every check passed and none are pending. `e2e` and `windows` were skipped by the workflow rules. `api-outcome` passed.

### Security: CLEAN
No new input surface, SQL is all Prisma-parameterised, and nothing runs in a shell. Regexes in `isTestFilePath` are linear (no ReDoS). The bench script's `file:` guard is weaker than its comment claims (Low, inline).

### Design / quality: GOOD
Pulling resolution into a pure `call-resolution.ts` is a real improvement, and so are the time-sliced yielder and the per-file transaction plus 500-row edge batching. The finding that `better-sqlite3` keeps awaited Prisma calls on the microtask queue, and that a same-process `/healthz` probe cannot see the stall, is a valuable catch.

### Performance: CLEAN
Ingest went from 458 s to 81 s. The overview's paged ranking walk is bounded.

### Tests: GAPS
New modules are covered (overview.ts 95.8% lines, matching the author's figure). Falsifiability is confirmed for #17. The #16 test cannot fail for 3 of its 4 yield sites. Its cost model spins when `create()` is *called* (inside the `.map` that builds the `$transaction` array), whereas real Prisma defers execution into `$transaction`. The real-DB check is the bench, which does not run in CI. Both are acceptable as long as they are stated.

### Documentation: UP TO DATE
Two fragments, `.changes/unreleased/16-…` and `17-…`, one per issue as CLAUDE.md requires. No CHANGELOG.md edit and no version bump. USER_GUIDE is updated.

### Deferred items acknowledged by the author (not re-raised as findings)
Same-package Java and cross-package Go receiver recall, Python `from x import`, CommonJS `require`, aliased namespace imports, and the legacy regex parser.

### Scope of this review
I read every changed production file in full and the new tests. I skimmed the mock-only additions to the nine `ingest-*-wiring` tests (`$transaction`/`createMany` shims, nothing more). I did not run the full monorepo `pnpm test` (CI did), did not re-run the bench, and did not exercise the running app.

server/src/lib/code-graph/ingest.ts
```diff
@@ -735,49 +761,59 @@ async function persistParsed(
...
+      language: file.language,
+      importedFiles: importTargets.get(file.filePath) ?? [],
+      runtimeImports: runtimeImports.get(file.filePath),
+    };
+
+    // Edges carry no ids anyone needs back, so they go in multi-row inserts:
+    // one statement (and one commit) per EDGE_INSERT_BATCH rows (#16).
+    let batch: Prisma.CodeEdgeCreateManyInput[] = [];
     for (const edge of file.edges) {
+      await maybeYield();
```
mgcronin, line 781: **Medium, high confidence, non-blocking. The #16 regression test guards only 1 of the 4 yield sites.**

I removed each `await maybeYield()` in turn and ran `ingest.test.ts -t "event loop" --retry=0`:
- L368 (parse loop): green
- L680 (pass-1 symbols): **red**
- L781 (this edge loop): green
- L1929 (rationale): green

The fixture has 300 files with 2 calls each, so the edge loop and parsing are too cheap to trip 500 ms. On the real workload (about 385k edges, 3.4k tree-sitter parses) these are the loops most likely to stall. Deleting this yield would pass CI and silently bring the blocked `/healthz` back.

Suggestion: add fixtures (or slow the parse/resolve step via an injected cost) so each site is load-bearing. Alternatively, inject the yielder and assert it is called from each phase.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,658 @@
...
+  }
+
+  if (receiver !== COMPLEX_RECEIVER) {
+    if (GLOBAL_RECEIVERS.has(receiver) || site.runtimeImports?.has(receiver)) return null;
+    const qualified = resolveQualifiedReceiver(receiver, name, site, index);
+    if (qualified) return qualified.id;
+  }
+
+  // Receiver of unknown type.
+  if (BUILTIN_METHOD_NAMES.has(name)) return null;
```
mgcronin, line 690: **Medium, medium confidence, non-blocking. The built-in guard discards import evidence for common domain method names.**

`BUILTIN_METHOD_NAMES` includes `get`, `set`, `add`, `update`, `delete`, `find`, `insert`, `remove`, `put`, `count`, `index`, `format`, `contains` and `size`. The check runs before the same-file and imported-file lookups, so `repo.update(x)` / `mapper.insert(o)` / `cache.get(k)` on an **imported** project class never binds. My probe confirms `repo.update` with `src/repo.ts::Repo::update` imported resolves to null.

On base, TS/Python imported-file resolution bound these. Impact-analysis caller sets for DAO/repository/mapper methods are exactly where this matters, and the recall audit in the PR covers only this TS repo.

Options:
- Narrow the list to names that are overwhelmingly built-in (`join`, `trim`, `map`, `filter`, `push`, `then`, …).
- Let an imported file that defines a *method* of that name count as evidence when the receiver identifier is itself an imported name.

At minimum, measure resolved `calls` edges on a Java/MyBatis fixture (e.g. jpetstore) before and after, and put the number in the PR next to the 76,519 → 60,723 figure.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,658 @@
...
+  "render",
+  "renderHook",
+  "screen",
+  "waitFor",
+  "within",
+  "fireEvent",
+  "act",
+  "cleanup",
+  // Python builtins
+  "print",
```
mgcronin: Low, high confidence. `GLOBAL_CALL_NAMES` applies to every language. Python/Go builtins (`filter`, `map`, `open`, `close`, `copy`, `delete`, `next`, `list`, `set`, `type`, `sum`, `min`, `max`, `all`, `any`, `make`, `append`) and RTL names (`render`, `act`, `cleanup`) therefore block a TS project's own unique top-level `open()`/`render()`/`copy()` from the step-3 fallback, and vice versa. Consider keying the set by `languageFamily(site.language)`.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,658 @@
...
+  "cy",
+  "assert",
+  "screen",
+  "userEvent",
+  "fireEvent",
+  "os",
+  "sys",
+  "re",
+  "json",
+  "path",
```
mgcronin: Low, medium confidence. `GLOBAL_RECEIVERS` also applies to every language and includes very common project module names: `path`, `errors`, `strings`, `json`, `re`, `os`. An in-project Go package `errors` (`internal/errors`), or TS `import * as errors from "./errors"`, then never binds through `resolveQualifiedReceiver`, even though `runtimeImports` already covers the TS `node:` cases precisely. Scoping by language, or relying on `runtimeImports` for JS, would avoid this.

server/scripts/bench-ingest.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+  parentPort.postMessage({ latencies, timeouts, errors });
+})();
+`;
+
+async function main(): Promise<void> {
+  const root = arg("root");
+  const dbUrl = process.env.DATABASE_URL ?? "";
+  if (!root) throw new Error("--root <dir> is required");
+  if (!dbUrl.startsWith("file:")) {
```
mgcronin, line 78: Low, high confidence. The guard says it refuses to touch a real database, but on this project the real dev database *is* a `file:` URL, so `DATABASE_URL=file:…/dev.db` passes. The script then creates a user and project and deletes them. That is harmless if it completes, but the delete is not in a `finally`, so an ingest that throws leaves `bench-*` rows behind. Consider refusing unless the path is under a temp dir, or unless an explicit `--scratch` flag is given, and move cleanup into `finally`.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,658 @@
...
+export interface ResolutionSite {
+  filePath: string;
+  language: string;
+  /** Project files this file's `imports` edges resolved to. */
+  importedFiles: string[];
+  /**
+   * Names this file imports from the runtime's standard library or a test
+   * framework ({@link isRuntimeOrTestModule}) — `import { join } from "node:path"`.
+   */
+  runtimeImports?: ReadonlySet<string>;
```
mgcronin, line 533: Low / informational (the author deferred this). `runtimeImports` is populated only from TS/JS `importedNames`. Python `from os.path import join` + bare `join()` still reaches the step-3 project-wide fallback, and the same goes for Go stdlib and Java static imports. Worth a follow-up issue so the deferral is tracked rather than just mentioned in the PR body.

mgcronin:
Review addressed in `6bcfa16`. Details and measurements are in the PR body under "Review follow-up".

1. **Blocking, fixed.**
   - `resolveBare` now checks `runtimeImports` first.
   - The same-file step skips methods, except in Java/C#.
   - The imported-file step only reaches top-level non-methods.
   - New tests: a fixture with both imports present (`node:path` `join` plus an imported `PathUtils::join` method, the self-edge case, and a `beforeEach` method), and 8 unit tests.
2. **Taken.** There is one event-loop test per phase: symbols, edges and rationale. Removing each yield turns only its own test red. The parse-loop yield stays green by construction, because the per-file `fs.readFile` await is already a turn.
3. **Taken, in a narrower form than either suggested option.** Collection names bind when the receiver is named after a class in an imported file.
   - Plain import evidence was measured and rejected: 153 `Map.get` calls became callers of `ConfigService::get`.
   - jpetstore-6 resolved calls: 792 (main), 828 (`f9e65e3`), 828 (`6bcfa16`).
   - A Java mapper fixture covers the `insert`/`update` shape.
4. and 5. **Taken.** Both lists are keyed by language family. An imported project module overrides a runtime receiver name. The Go in-project `errors` package case still needs Go import resolution, which is in #65.
6. **Taken.** The bench refuses a database that already holds any project, and cleans up in a `finally`.
7. **Filed** as #65.

server/src/lib/code-graph/call-resolution.ts
```diff
@@ -0,0 +1,762 @@
...
+    return parts.length === 2 && parts[0].toLowerCase() === want;
+  });
+  return hits.length === 1 ? hits[0] : null;
+}
+
+function resolveBare(name: string, site: ResolutionSite, index: ResolutionIndex): string | null {
+  // 0. A name this file imports from the runtime or a test framework
+  //    (`import { join } from "node:path"`) is by definition not a project
+  //    symbol — not even a same-file or imported one of the same name.
+  if (site.runtimeImports?.has(name)) return null;
```
mgcronin, line 728: For information: I checked this by running it. With the runtime-import guard first and the bare/top-level indices in steps 1 and 2, every probe from the previous round now resolves to null, and a legitimate imported top-level function still binds. Resolved.

mgcronin (commented):
Recommendation: merge — no blocking findings

This is automated analysis published as a COMMENT, not a human sign-off. The reviewing gh identity is also the PR author, so GitHub cannot record an approval from it. A human owns the merge decision.

## Re-review (cycle 2) of 6bcfa16

### Prior blocking finding 1 (bare calls bound in steps 1 and 2, before the runtime-import guard): FIXED, verified by running it
The runtime-import check now comes first in `resolveBare`. Step 1 uses the non-method index, except in Java and C#. Step 2 uses a top-level-only index.

I ran my original probes against the resolver on the PR head. All 7 now behave correctly:
- `join` imported from node:path, next to an imported `PathUtils::join` method: null.
- The same, next to an imported top-level `join`: null.
- A bare `join` with no runtime import still binds to the imported top-level function.
- `beforeEach` imported from vitest, next to a helper: null.
- A globals-mode `beforeEach`, next to an imported helper method: null.
- `repo.update()` with `Repo` imported: binds.
- `analyses.update()`: does not bind.

**The new tests can fail.** With only `call-resolution.ts` put back to f9e65e3, 16 tests fail across `call-resolution.test.ts` and `ingest.test.ts`. That includes both ingest-level `(#64 review)` fixtures.

### Prior non-blocking findings
- **2 (the event-loop test covered one yield only): addressed.** I deleted the edge-loop yield (ingest.ts:781): only the pass-2 test went red. I deleted the rationale yield (ingest.ts:1929): only the rationale test went red. The other tests stayed green both times.
- **3 (`get` / `update` / `insert` lost their edges): addressed.** These names now need the receiver to be named after an imported class. My probe and the author's Java mapper fixtures both confirm it.
- **4 and 5 (built-in and namespace lists not scoped by language): addressed.** The lists are now per language, and importing a project module of the same name overrides them. The Go package case is deferred to #65.
- **6 (bench guard): addressed.** The bench refuses a database that already holds a project, and cleans up in a `finally`.
- **7:** tracked as #65.

### Gate (run in a throwaway worktree at 6bcfa16, since removed)
- `pnpm install --frozen-lockfile --prod=false`, then the `@metis/shared` build, then `pnpm typecheck` in server: exit 0.
- eslint on the 4 files that changed since f9e65e3: clean.
- vitest on code-graph, ingest-*, impact-analysis and docs-gen: 27 files, 406 tests passed.
- CI on 6bcfa16: all checks passed, none pending. `api-outcome` passed; `e2e` and `windows` were skipped by the workflow rules.
- Not re-run: the full monorepo `pnpm test` (CI covered it) and the benchmark numbers in the PR body.

### New findings
None blocking. The two inline comments are for information only.

server/src/lib/code-graph/ingest.ts
mgcronin, line 781: For information: I removed this yield, and separately the one at L1929. Each time only that phase’s event-loop test went red, so the earlier gap is closed.

Issue #70, opened 2026-09-22
Impact analyses are invisible on their project's page until the first item is written (follow-up to #61)


mgcronin:
_Found while fixing #61._

**An impact analysis is invisible on its project's page until it has written its first item.**

The projects a run was started for are not persisted: `triggerImpactAnalysis` (`server/src/lib/impact-analysis/impact-analysis-engine.ts`) creates the `ImpactAnalysis` row with no project reference and passes `projectIds` only to the background executor. `listImpactAnalyses` derives a run's projects from its `ImpactItem` rows, so a run that is `pending`/`running`, failed before any item, or extracted zero changes has no projects.

Consequences:

1. `/projects/:id/impact` (#61) cannot show a run that is still in progress, or one that failed early.
2. In `GET /api/impact-analyses`, the access filter lets a run with no items through for every non-admin (`projectIds.length === 0 || …`), so a member sees other users' in-progress runs on projects they cannot access (status and timestamps only).

## Suggested fix

Persist the selected project ids on the run (a join table or column, with a migration), derive list membership and access from them, and keep the item-derived ids for detail views.

## Acceptance criteria

- [ ] A running impact analysis appears on each selected project's Impact Analysis page
- [ ] A member does not see another user's in-progress run on a project they cannot access

Issue #69, opened 2026-09-22
Documents page polls forever and shows 'processing' for quarantined documents (follow-up to #66)


mgcronin:
_Found while fixing #66._

**The project Documents page treats quarantined documents as still ingesting.**

A document held in quarantine keeps `status = processing` (`indexState = quarantined`). #66 fixed the Overview; the Documents page (`ui/src/app/(authed)/projects/[id]/documents/page.tsx`) still

- polls `GET /documents` every 3 s for as long as any quarantined document exists (`INGESTING_STATUSES` includes `processing`), and
- lists it as `processing · N chunks`, with nothing saying it is waiting for review.

## Suggested fix

Use `isDocumentIngesting` / `isDocumentAwaitingReview` from `ui/src/lib/project-pipeline.ts` (added by #66) for the poll, and label quarantined rows "awaiting review" with a link to `/projects/:id/settings#quarantine`.

## Acceptance criteria

- [ ] With only quarantined documents, the page stops polling
- [ ] A quarantined row reads as awaiting review and links to the quarantine queue

Issue #68, opened 2026-09-22
Server logger writes an Error in log metadata as {} — message and stack are lost


mgcronin:
_Found while fixing #52._

**The server logger records an `Error` passed in log metadata as `{}`.**

`redact()` in `server/src/lib/logger.ts` rebuilds every metadata object from `Object.entries(...)`. An `Error`'s `message` and `stack` are not own enumerable properties, so `log.error("…", { err })` is written as `"err":{}`. Measured:

```
redactInfo({ level: "error", message: "m", err: new Error("secret detail"), docId: "d" })
→ {"level":"error","message":"m","err":{},"docId":"d"}
```

`winston.format.errors({ stack: true })` only handles an `Error` passed as the log *message* or the top-level info object, not one nested in metadata. Every call site that logs `{ err }` (there are many under `server/src`) records no detail at all.

#52 worked around it at the one call site it touched (the doc-generation failure log now logs `error` and `stack` as strings).

## Suggested fix

Teach `redact()` to serialise an `Error` as `{ name, message, stack, ...own enumerable props }` before recursing, so key-based redaction still applies to its properties. Add a test through `redactInfo`, the path production uses.

## Acceptance criteria

- [ ] `log.error("x", { err: new Error("boom") })` writes the message and the stack
- [ ] A sensitive-named own property on an error (e.g. `err.token`) is still redacted

Issue #67, opened 2026-09-22
Degraded generated docs still expose raw exception text in section warnings (follow-up to #52)


mgcronin:
_Found while fixing #52._

**A degraded generated document's section warnings still carry raw exception text.**

`sectionFailedWarning(section, detail)` (`server/src/lib/docs-gen/grounding/degraded-warnings.ts`) appends up to 300 characters of `detail` to the warning message, and `holistic-synthesizer.ts` passes `String(err)` for a section whose generation threw. Those warnings are persisted in the `warnings` column and returned by `GET /projects/:projectId/docs/:docId`; the UI's `DegradedWarningsBanner` renders them.

That is the same exposure #52 closed for a *failed* document's `errorMessage` — provider response bodies, paths or SQL text reaching the browser — on the *degraded* path.

## Suggested fix

Map the section's error through `generationFailureMessage` (`server/src/lib/docs-gen/generation-failure-message.ts`, added by #52) before building the warning, log the raw error server-side, and sanitise persisted `section-failed` warnings on read the way #52 does for `errorMessage`.

## Acceptance criteria

- [ ] A section that throws `Error("secret detail")` produces a warning whose message does not contain it, read back through `GET /docs/:docId`
- [ ] A provider 402 in a section stays recognisable as a balance problem

PR #71, opened 2026-09-22
fix: Resolve #52, #61, #66 — safe doc-failure reasons; project impact runs; quarantined docs await review

### LINKED ISSUE #52: Generated-doc detail API returns raw exception text for failed generations

### LINKED ISSUE #61: Project-scoped Impact Analysis page cannot list this project's runs — summary drops projectIds
Follow-up to #28 (epic #26).

## Problem

#28 added a project-scoped **Analyze → Impact Analysis** page (`/projects/[id]/impact`). It can start a new impact analysis with the project pre-selected (`/impact-analyses/new?projectId=…`), but it cannot list *this project's* past impact analyses: `GET /api/impact-analyses` returns `ImpactAnalysisSummary` rows (`packages/shared/src/impact.ts`) with only `projectCount`, not the project ids. `listImpactAnalyses` in `server/src/lib/impact-analysis/impact-analysis-read.ts` already computes `projectIds` per row and then drops them.

#28 was scoped to `ui/`, so the page links to the global list instead.

## Proposed

- Add `projectIds: string[]` to `ImpactAnalysisSummary` (read-only; already computed and already access-filtered).
- Optionally accept `?projectId=` on `GET /api/impact-analyses` so the filter runs server-side.
- Render this project's runs on `/projects/[id]/impact`.

## Acceptance criteria

- [ ] `/projects/[id]/impact` lists only impact analyses that include this project
- [ ] A user without access to a run's other projects sees no more than they do today on `/impact-analyses`

### LINKED ISSUE #66: Project Overview shows 'Ingesting' for quarantined generated docs (follow-up to #63)
**The new project Overview (#63) shows "Ingesting — at least 3 documents processing" when nothing is ingesting.** The three documents are generated documentation that METIS published back into its knowledge base and quarantined for review.

Observed on a local install after #63 merged:

| filename | status | indexState | chunkCount | processedAt |
|---|---|---|---|---|
| `generated-doc-….md` (Architecture) | `processing` | `quarantined` | 26 | set at upload |
| `generated-doc-….md` (Business Requirements) | `processing` | `quarantined` | 58 | set at upload |
| `generated-doc-….md` (degraded run) | `processing` | `quarantined` | 1 | set at upload |

Each was chunked and processed at upload; it is waiting for review, not for ingestion. The Overview's Ingest card counts `status = processing` rows as in-flight ingestion, so it spins indefinitely for any project with quarantined generated docs.

## Two defects, pick which to fix (or both)

1. **UI** — the Overview's Ingest stage should treat `indexState = quarantined` as "awaiting review", not "ingesting", and ideally surface it under Review with a link to the quarantine queue.
2. **Data model** — a document whose processing finished and which is only held in quarantine arguably should not keep `status = processing`. Check what else reads `status` (the documents list, the Workbench) before changing it.

## Acceptance criteria

- [ ] A project whose only non-ready documents are quarantined does not show "Ingesting"
- [ ] Quarantined documents are counted and linked as "awaiting review"
- [ ] A test with a quarantined generated document asserts the Ingest card is idle

mgcronin:
Fallout from today's merges, bundled as requested: one PR for three issues.

Closes #52
Closes #61
Closes #66

## #52: a failed generation's raw exception text no longer reaches the client

- `generateDocumentAsync` used to store `String(err)`. It now stores `generationFailureMessage(err)`, which returns one of a fixed set of messages and never echoes the error: the #53 restart message (passed through unchanged), **"402 Insufficient Balance"** from a provider, the project's monthly token budget (`BudgetExceededError`), a 429 rate limit, rejected credentials (401/403), or a generic failure. The error is classified by its numeric `status` and by known phrases. A bare number in the message does not count as a status: `"parsed 402 tables"` falls through to the generic message.
- `GET /docs/:docId` and `PATCH /docs/:docId` both returned the whole row. Both now run a `failed` row's `errorMessage` through `publicGenerationErrorMessage`, so **rows written before this change are cleaned up when they are read**. A `degraded` row's legacy warning JSON is passed through untouched.
- **Defect found: the server log was not keeping the detail either.** `redact()` in `server/src/lib/logger.ts` rebuilds metadata from `Object.entries`. An `Error`'s `message` and `stack` are not enumerable, so `log.error("…", { err, docId })` wrote `"err":{}`. I confirmed this with a probe through `redactInfo`. Until now the database row was the only place the error was recorded. The call site now logs `error` and `stack` as strings. The logger-wide fix is filed as #68.

## #61: the project Impact Analysis page lists its own runs

- `ImpactAnalysisSummary.projectIds` lists only the run's projects the caller can access (all of them for an admin). `projectCount` is unchanged, so a member sees no more than before.
- `GET /api/impact-analyses?projectId=` applies the filter in the Prisma query (`items: { some: { projectId } }`) before the 100-row `take`. **A client-side filter would have been wrong-layer:** once 100 newer runs of other projects exist, a project's older runs would drop out of the list silently. A project the caller cannot access returns `[]`.
- `/projects/[id]/impact` renders the runs through the table now shared with `/impact-analyses` (`ImpactAnalysesTable`). The page also checks `projectIds`, so a run from another project never shows even if the server filter is lost. The query key sits under `list()`, so the existing invalidation after a run finishes also refreshes it.

## #66: quarantined documents show as awaiting review, not ingesting

- UI-only. **`Document.status` semantics are unchanged.** Other code reads `status = processing` for quarantined rows: the Documents list, the approve/reject flow in `server/src/lib/rag/quarantine.ts`, and `knowledge-service.ts`. The UI now tells the two apart with `indexState`.
- `isDocumentIngesting` / `isDocumentAwaitingReview` in `ui/src/lib/project-pipeline.ts`:
  - The Ingest stage no longer counts quarantined documents as processing. It adds "N documents awaiting review" instead.
  - The Overview's documents poll, which polled forever, also stops for them.
  - The **Review** stage counts them, with a link to the quarantine queue (`/projects/:id/settings#quarantine`; the panel now has `id="quarantine"`). Requirements waiting for review keep the primary action, and quarantine becomes a secondary one (`PipelineStage.secondaryAction`).

## Tests, and the mutations that prove they can fail

Every mutation below was applied, the named tests went red, and the code was restored.

| # | Mutation | Red |
|---|---|---|
| 52-1 | Write `errorMessage: String(err)` again (the original code) | e2e write→read test: stored message is not generic |
| 52-2 | GET returns `doc.errorMessage` unsanitised | legacy-row test (raw text in body) |
| 52-3 | PATCH returns `updated.errorMessage` unsanitised | PATCH test |
| 52-4 | Log `{ err, docId }` again | "logs the detail server-side" (logged `{}`) |
| 52-5 | Drop the 402 branch | 402 unit + legacy re-classification tests |
| 52-6 | `publicGenerationErrorMessage` passes text through | 3 tests (unit, route, e2e) |
| 52-7 | Drop the safe-message passthrough | restart-message tests |
| 61-1 | Drop `projectIds` from the summary | 2 list tests |
| 61-2 | Return unfiltered `projectIds` to a member | "names only the projects a member can access" |
| 61-3 | Drop the query `where` (filter after `take`) | "filters to one project in the query, ahead of the limit" |
| 61-4 | Drop the route's access check | "lists nothing for a project the caller cannot access" |
| 61-5 | Page drops its `projectIds` filter | "never lists a run that does not include this project" |
| 61-6 | Hook ignores `projectId` | hook test + page "asks the server" |
| 61-7 | Project key outside the `list()` prefix | invalidation test |
| 66-1 | `isDocumentIngesting` ignores quarantine (the original behaviour) | 4 tests incl. the issue's three-row repro and polling |
| 66-2 | Overview poll keyed on `status` again | polling test |
| 66-3 | Review stage ignores quarantine | 3 tests |
| 66-4 | Quarantine never the primary action | 2 tests |
| 66-5 | Secondary action not rendered | component test |
| 66-6 | Ingest omits "awaiting review" | 2 tests |
| 66-7 | Remove `id="quarantine"` from the panel | anchor test |

The #52 end-to-end test drives the real `generateDocumentAsync` against the stateful in-memory table in `generated-doc-interrupted-generation.test.ts`. It rejects the synthesizer with a secret-bearing error and reads the row back through `GET /docs/:docId`, the same path the UI uses.

## Coverage (touched files, targeted runs)

|---|---|---|
| `server/src/lib/docs-gen/generation-failure-message.ts` (new) | — → 100% | — → 95.7% |
| `server/src/lib/impact-analysis/impact-analysis-read.ts` | 94.8% → 99.5% | 77.0% → 84.1% |
| `server/src/routes/generated-docs.ts` | 92.6% → 92.6% | 73.8% → 73.5% |
| `server/src/routes/impact-analysis.ts` | 49.5% → 50.2% (pre-existing; every new line is covered) | 42.7% → 46.4% |
| `ui/src/lib/project-pipeline.ts` | 100% | 94.7% |
| `ui/src/components/projects/pipeline-overview.tsx` | 100% | 87.0% |
| `ui/src/components/impact/impact-analyses-table.tsx` (new) | 100% | 100% |
| `ui/src/app/(authed)/projects/[id]/impact/page.tsx` | 100% | 90% |

## Gate (run locally, from CLAUDE.md)

- `pnpm lint`: exit 0.
- `pnpm typecheck`: exit 0.
- `pnpm test`: **exit 1, and not because of this change.**
  - `scripts/lib/gate-cwd-anchoring.test.mjs` (the company-identifier scan run twice) hit its 5 s timeout under the monorepo fan-out, on both runs. It passes 9/9 on its own. This PR changes nothing under `scripts/`.
  - `@metis/server` + `@metis/ui` pass on their own: 1118 files / 17556 tests and 360 files / 3373 tests. shared, ui-kit, copilot-svc and embeddings-svc also pass.
  - During one doubled fan-out run, `tests/socket.test.ts` also timed out connecting. It passes 7/7 on its own, with or without this change.
- `pnpm changelog:verify`: 3 fragments contributed; every fragment parses.

## Security

- **OWASP pass over the diff.** `projectId` from the query string reaches Prisma only as a parameterised `where`, and access is checked before the query. No shell, no raw SQL, no new secrets.
- The stack now written to the server log stays server-side. It passes through the existing key-based redaction.
- No dependencies changed, so no audit was run locally. CI's Dependency audit job covers it.

## Deferred (filed)

- #67: a degraded document's `section-failed` warnings still include `String(err)` (`holistic-synthesizer.ts`), and the UI renders them. Same leak as #52, on the degraded path.
- #68: the logger writes an `Error` in metadata as `{}` at every `{ err }` call site.
- #69: the Documents page still polls every 3 s and shows `processing` for quarantined documents.
- #70: an impact run's projects are not persisted, so an in-progress or early-failed run is invisible on its project's page. Separately, zero-item runs pass the access filter for every member. Fixing either needs a schema change, which is out of scope for a read-only fix.

## Not done

- No new umbrella issue. The three issues this PR closes already record why the work was done.
- No e2e spec. The e2e job belongs to another agent's concurrent work (#62).


server/src/lib/impact-analysis/impact-analysis-read.ts
```diff
@@ -842,13 +842,21 @@ export function computeSharedTableImpacts(items: ImpactItemView[]): SharedTableI
...
+ * access; `projectCount` still counts them all, exactly as before.
+ */
 export async function listImpactAnalyses(
-  opts: { accessibleProjectIds?: string[] | null; limit?: number } = {},
+  opts: { accessibleProjectIds?: string[] | null; limit?: number; projectId?: string } = {},
   prisma?: ReadPrisma,
 ): Promise<ImpactAnalysisSummary[]> {
   const db = resolvePrisma(prisma);
   const rows = await db.impactAnalysis.findMany({
+    ...(opts.projectId ? { where: { items: { some: { projectId: opts.projectId } } } } : {}),
```
mgcronin, line 859: The project filter has to be in the query, not applied afterwards: `take` caps the list at the newest 100 runs, so filtering the result would silently drop a project's older runs once other projects have 100 newer ones. Pinned by `impact-analysis-list.test.ts` › 'filters to one project in the query, ahead of the limit'.

server/src/routes/generated-docs.ts
```diff
@@ -1142,11 +1153,24 @@ export async function generateDocumentAsync(
       healthStatus === "degraded" ? "Generated with warnings" : "Documentation ready",
     );
   } catch (err) {
-    log.error("Document generation failed", { err, docId });
+    // #52 — the only record of the raw error. Logged as strings: the logger's
+    // redaction pass copies own enumerable properties, and an Error's
+    // `message` and `stack` are not, so `{ err }` logged `err: {}`.
+    log.error("Document generation failed", {
```
mgcronin, line 1159: `{ err }` was being logged as `"err":{}`: the redaction pass in `logger.ts` copies only own enumerable properties, and an Error's message and stack are not. Now that the row no longer stores the raw text, this log line is the only record of it, so it logs strings. The logger-wide fix is #68.

mgcronin (commented):
## Code Review — PR #71 (automated analysis, not a human sign-off)

Recommendation: **merge with follow-ups — 0 blocking findings, 6 non-blocking**

The reviewing `gh` identity (`mgcronin`) is also the PR author, so GitHub cannot record an approval or a change request from it. This is published as a comment; the merge decision stays with a human.

### What I verified rather than accepted

- **CI**: full check set present on a plain re-read, **0 pending, 0 failing** (`api` pass 25m54s, `api-outcome` pass, `ui`, `postgres-adapter`, `Semgrep`, `CodeQL`, `Dependency audit`, `changelog`, `sql-lineage`, `generative-e2e`, `postgres-migrate-deploy` all pass; `e2e`/`windows` skipping as on `main`).
- **Gate**, in a fresh worktree off `refs/pull/71/head` (`pnpm install --frozen-lockfile --prod=false` → `pnpm --filter @metis/shared build` → `pnpm typecheck`): **exit 0**. `eslint` over the 8 changed source files: **clean**. I did not re-run the whole `pnpm test` fan-out — CI's `ui` and `api` jobs cover it and the author's two timeout diagnoses are consistent with CLAUDE.md's §3 fan-out-starvation reading.
- **Targeted tests**: server 5 files / 110 tests pass; ui 18 files / 273 tests pass.
- **Mutations re-run independently** (production code reverted, tests watched, then restored):
  | Mutation | Result |
  |---|---|
  | `errorMessage: String(err)` restored in `generateDocumentAsync` | RED — `generated-doc-interrupted-generation.test.ts` |
  | `GET /:docId` returns `doc.errorMessage` unsanitised | RED — legacy-row test |
  | `where: { items: { some: { projectId } } }` removed | RED — "filters to one project in the query, ahead of the limit" |
  | `isDocumentIngesting` ignores quarantine | RED — 4 tests incl. the issue's three-row repro |
  All four claims hold. The tests read back through the real consumer path (`GET /projects/:id/docs/:docId`, `derivePipelineStages`, `apiFetch`), not against the object they just built.
- **The #66 fix is not a no-op.** `GET /projects/:projectId/documents` (`server/src/routes/documents.ts:328`) does a `findMany` with **no `select`**, so the whole Prisma row — `indexState` included — is already on the wire. The new UI field has a real producer.
- **`ImpactItem` is indexed on `projectId`** (`server/prisma/schema.prisma`), so `items: { some: { projectId } }` is not a scan.

### Requirements: PASS (with two partials)

| Issue | Criterion | Status |
|---|---|---|
| #52 | raw exception text no longer stored | PASS |
| #52 | raw exception text no longer returned by `GET`/`PATCH` | PASS for `doc.errorMessage`; see finding 1 and 2 |
| #52 | restart message + 402 stay recognisable **to the user** | PARTIAL — see finding 3 |
| #52 | detail logged server-side | PASS, and the `{ err }` → `{}` logger defect found on the way is a genuinely good catch |
| #61 | `/projects/[id]/impact` lists only runs including this project | PASS |
| #61 | a member sees no more than on `/impact-analyses` | PASS (`projectCount` unchanged, `projectIds` access-filtered) |
| #61 | read-only change | PASS |
| #66 | quarantine-only project does not show "Ingesting" | PASS |
| #66 | counted and linked as awaiting review | PASS |
| #66 | other `Document.status` readers checked first | PASS — data model untouched |

### Security: FINDINGS (none introduced by this PR)

| Severity | Count | Where |
|---|---|---|
| High | 0 | — |
| Medium | 2 | findings 1 and 2 below |
| Low | 2 | findings 4 and 5 |

`projectId` reaches Prisma only as a parameterised `where` behind an access check; no shell, no raw SQL, no new secrets, no dependency changes. The `?projectId=` inaccessible-project answer (`[]`) is indistinguishable from "no runs", which is the right call.

### Tests: SUFFICIENT · Documentation: UP TO DATE

Three `.changes/unreleased/` fragments, no `CHANGELOG.md` edit, no version bump — per CLAUDE.md. `docs/ARCHITECTURE.md` updated for both the new module and the new query parameter. One wording caveat in finding 2.

### Scope of this review

Read all 30 changed files plus the unchanged readers they depend on (`documents.ts`, `quarantine.ts`, `incremental.ts`, `task-store.ts`, `task-queue.ts`, `documentation/page.tsx`, `schema.prisma`). Ran typecheck, scoped lint, the targeted suites and four independent mutations. Did not exercise the running application, and did not touch the concurrent agents' files.

Six findings follow as inline comments. None of them is a regression: every one is either a residual gap in a pre-existing path or a scope question, which is why the recommendation is merge-with-follow-ups rather than hold.

server/src/lib/docs-gen/generation-failure-message.ts
```diff
@@ -0,0 +1,125 @@
...
+ * `failed` row's message is an error reason; a legacy `degraded` row may still
+ * carry pre-#252 warning JSON the UI parses, which METIS wrote itself, so it
+ * passes through.
+ */
+export function publicGenerationErrorMessage(
+  status: string,
+  errorMessage: string | null | undefined,
+): string | null {
+  if (errorMessage == null) return null;
+  if (status !== "failed") return errorMessage;
```
mgcronin, line 123: **Medium (high confidence) — the read-time sanitiser is opt-in on one status, so a legacy raw message survives into the `generating` window.**

`if (status !== "failed") return errorMessage;` passes through for `pending`, `generating`, `ready` and `degraded`. Only `degraded` actually needs it (the pre-#252 warning JSON your docstring describes). The other three are reachable with a stale raw message:

- `checkIncrementalRegeneration` selects `status: { in: ["ready","degraded","failed","generating"] }` with `autoUpdate: true` (`server/src/lib/docs-gen/incremental.ts:29`) — so a **failed** document is picked up for automatic regeneration;
- the claim at `server/src/routes/generated-docs.ts:716` writes `data: { status: "generating", codeGraphHash: claim }` and **does not clear `errorMessage`** (the manual `/regenerate` path does, at line 540);
- `GET /projects/:id/docs/:docId` then returns the pre-#52 raw string verbatim for the whole generating window.

I measured it rather than inferring it:

```
publicGenerationErrorMessage("generating", 'Error: secret detail at /srv/metis/server/src/x.ts:9')
=> 'Error: secret detail at /srv/metis/server/src/x.ts:9'
```

Deny-by-default is a one-line change and removes the whole class:

```ts
if (status === "degraded") return errorMessage;   // legacy warning JSON METIS wrote itself
return generationFailureMessage(errorMessage);
```

Narrow in practice (needs a pre-#52 row *and* `autoUpdate`), so I am not calling it blocking — but it is a gap in the guard this PR added, in the exact route the security issue names, and it costs one line.

server/src/routes/generated-docs.ts
```diff
@@ -400,6 +404,8 @@ export function generatedDocsRouter(): Router {
     res.json({
       data: {
         ...doc,
+        // #52 — never the raw exception text a pre-#52 row may still hold.
+        errorMessage: publicGenerationErrorMessage(doc.status, doc.errorMessage),
```
mgcronin, line 408: **Medium (high confidence) — the same response body still carries raw exception text, one field over.**

`doc.errorMessage` is now safe, but a few lines below (413-418, and the list route at 330-337) `indexing.errorMessage` is copied verbatim from `Document.errorMessage` / `Task.errorMessage`, and both are written raw:

- `server/src/lib/scheduler/task-store.ts:143` — `markFailed` persists `error.slice(0, 4000)`, where `error` is `err instanceof Error ? err.message : String(err)` (`task-queue.ts:344`). That row is the generated-doc publication outbox task, surfaced through `unpublishedIndex(outbox)`.
- `server/src/lib/rag/quarantine.ts:606-607` — a failed reconciliation writes `error instanceof Error ? error.message : String(error)` to `Document.errorMessage`.

And the UI *renders* it: `ui/src/app/(authed)/projects/[id]/documentation/page.tsx:1099` — `indexing.errorMessage?.trim() || "Indexing was rejected."`.

This is **pre-existing, not a regression** — the PR strictly reduces exposure. Two asks:

1. File a fifth follow-up alongside #67/#68/#69/#70 for `indexing.errorMessage`, since it is the closest sibling of the field #52 is about and lands in the same payload.
2. Narrow the changelog fragment. `.changes/unreleased/52-doc-failure-message.md` currently reads "A failed documentation generation no longer returns the server's raw exception text from `GET /projects/:id/docs/:docId`" — true of `errorMessage`, not of the response. "no longer returns the server's raw exception text as the document's failure reason" would be exactly true.

server/src/lib/docs-gen/generation-failure-message.ts
```diff
@@ -0,0 +1,125 @@
...
+ * a few well-known phrases in its message — never echoed. The raw error stays
+ * in the server log, where `generateDocumentAsync` already writes it.
+ */
+import { GENERATION_INTERRUPTED_MESSAGE } from "./interrupted-generations.js";
+
+export const GENERATION_FAILED_MESSAGE =
+  "Document generation failed. The details are in the server log; regenerate the document to try again.";
+
+export const GENERATION_PROVIDER_BALANCE_MESSAGE =
+  "The AI provider refused the request: 402 Insufficient Balance. Top up the provider account, then regenerate the document.";
```
mgcronin, line 20: **Medium (high confidence) — five of the six messages have no consumer, so the "recognisable to the user" half of #52 is unproven.**

`FailedGenerationBanner` (`ui/src/app/(authed)/projects/[id]/documentation/page.tsx:1234-1257`) takes `interrupted`, `regenerating` and `onRegenerate` — **not `errorMessage`** — and renders one of two fixed strings. The `interrupted` boolean itself is computed from the *raw* `doc.errorMessage` at `generated-docs.ts:410`, not from your classified value. `errorMessage` is passed to the UI only for `degraded` documents (line 474), which never take this path.

So `GENERATION_PROVIDER_BALANCE_MESSAGE`, `GENERATION_BUDGET_EXCEEDED_MESSAGE`, `GENERATION_PROVIDER_RATE_LIMITED_MESSAGE` and `GENERATION_PROVIDER_AUTH_MESSAGE` are computed, persisted and served — and nothing in the product reads them. The three status regexes and four classification branches are the largest new surface in this PR, and their only exercise is their own unit tests. That is the "state a control writes and nothing renders from" shape: it passes every test it has and changes nothing a user sees.

Two honest options, either fine:

- **Consume it** — one line in the banner (`{errorMessage}` under the headline, gated on it being one of `SAFE_MESSAGES`), which is what makes "402 Insufficient Balance recognisable to the user" actually true and is what makes the classifier's defects surface;
- **Or say so** — note in the PR and the changelog that the classification is API-only for now, with a follow-up for the banner.

What should not stand is the changelog's current implication that a user sees the provider reason. Credit where due: the classifier itself is careful — the "a bare number is not a status" rule and the `openai returned 429` case you added after a surviving mutation are both exactly the right instinct.

server/src/routes/impact-analysis.ts
```diff
@@ -733,14 +733,23 @@ export function impactAnalysisRouter(deps: UsageClassificationRouterDeps = {}):
...
+  // to runs that include that project; a project the caller cannot access
+  // lists nothing, the same answer as a project with no runs.
   r.get("/", requirePermission("analysis.read"), async (req, res, next) => {
     try {
       const actor = actorOf(req);
       const accessibleProjectIds = isAdminActor(actor)
         ? null
         : await listAccessibleProjectIds(actor);
-      const rows = await listImpactAnalyses({ accessibleProjectIds });
+      const projectId = typeof req.query.projectId === "string" ? req.query.projectId : "";
```
mgcronin, line 745: **Low — a repeated `?projectId=` silently returns the *global* list.**

Express parses `?projectId=a&projectId=b` into an array, so `typeof req.query.projectId === "string"` is false and `projectId` falls back to `""` — which skips both the access check and the query filter, and answers with every run the caller can see. Not a security hole (the `accessibleProjectIds` filter in `listImpactAnalyses` still applies), but it is a silently wrong answer where a `400` would be the honest one.

```ts
const raw = req.query.projectId;
if (raw !== undefined && typeof raw !== "string") {
  throw new AppError(400, "VALIDATION_ERROR", "projectId must be a single value");
}
```

nit, same line: `""` as the "absent" sentinel means an explicit `?projectId=` (empty) is treated as absent too. That is probably what you want, but it is worth one word of comment since the access check is skipped on that branch.

ui/src/lib/project-pipeline.ts
```diff
@@ -85,6 +90,25 @@ export const FIRST_RUN_STEPS: readonly PipelineStageId[] = [
 ];
 
 const RUNNING_DOC_STATUSES = new Set(["pending", "queued", "processing"]);
+
+/**
+ * #66 — a document held in quarantine: processed and chunked, waiting for a
+ * reviewer. It keeps `status = processing` (the documents list and the
+ * approve/reject flow read it that way), so status alone reads as ingesting.
+ */
+export function isDocumentAwaitingReview(d: { indexState?: string | null }): boolean {
```
mgcronin, line 99: **Low — `isDocumentAwaitingReview` ignores `status`, so a rejected/failed document can be counted as awaiting review.**

`rejectDocument` (`server/src/lib/rag/quarantine.ts:646-652`) sets `indexState: "rejected"`, so the common reject path is clean. But any row that reaches `status: "failed"` while `indexState` is still `"quarantined"` would be counted **twice**: once in Ingest's `failed` tally and once in Review's "awaiting review", with a Review action pointing at a queue that `listQuarantine` may or may not still list it in.

`isDocumentIngesting` already pairs `status` with `indexState`; doing the same here keeps the two predicates symmetric:

```ts
return d.indexState === "quarantined" && d.status !== "failed";
```

Low confidence that this is reachable today — I did not find a writer that produces the combination — which is why it is a Low rather than a defect claim. Worth the guard anyway, since it costs nothing and the two functions currently disagree about whether `status` matters.

ui/src/components/projects/pipeline-overview.tsx
```diff
@@ -105,9 +118,8 @@ export function ProjectPipelineOverview({ projectId }: { projectId: string }) {
     queryKey: [...queryKeys.documents.forProject(projectId), "pipeline"],
     queryFn: () => documentsApi.list(projectId, { limit: DOCUMENT_PAGE }),
     refetchInterval: (q) =>
-      q.state.data?.items.some((d) => ["pending", "queued", "processing"].includes(d.status))
-        ? LIVE_POLL_MS
-        : false,
+      // #66 — a quarantined document is waiting for review, not ingesting.
+      q.state.data?.items.some(isDocumentIngesting) ? LIVE_POLL_MS : false,
```
mgcronin, line 122: **Low — the forever-poll survives for `reconciling`.**

Good fix, and the poll predicate now sharing `isDocumentIngesting` with the Ingest card is the right structure — one definition, two readers.

One residual: `server/src/lib/rag/quarantine.ts:547` and `:606` leave a document at `indexState: "reconciling", status: "processing"` during post-approval cleanup, and `:607` leaves it at `reconciling` + `failed` when that cleanup throws. A document stuck in the first of those still matches `RUNNING_DOC_STATUSES`, so this query still polls every `LIVE_POLL_MS` indefinitely — the exact symptom #66 reports, one state over.

`reconciling` is genuinely transient (unlike `quarantined`), so excluding it from `isDocumentIngesting` would be wrong. A bounded poll — stop after N minutes with no status change — would close the shape for good. Out of scope here; worth a line in #69, which already owns the Documents page's 3-second forever-poll.

Issue #73, opened 2026-09-22
No test covers coverage-service passing its cost guard to the judge


mgcronin:
`runCoverageScoring` (`server/src/lib/testcoverage/coverage-service.ts`) passes the run's `CoverageCostTracker` to `judgeAmbiguous` as `cost`. The judge uses it to record each batch and stop between batches when the budget is reached (#883, #43).

No test notices if that argument is removed. I deleted `cost` from the `judgeAmbiguous(...)` call and ran the test-coverage and finops suites: all 139 tests passed. Without `cost`, the judge falls back to the legacy path. It records one total at the end through the global `TokenTracker`, and it never checks the budget. So a run could make every judge call past the cap, and the service tests would stay green.

`judge.test.ts` covers the guard inside `judgeAmbiguous`. Nothing covers the service passing the guard in.

Found while mutation-testing the PR for #57/#58.

## Acceptance criteria
- [ ] A `coverage-service` test with AMBIGUOUS pairs and a small budget shows the judge phase stopping part-way.
- [ ] That test goes red when `cost` is removed from the `judgeAmbiguous` call.

Issue #72, opened 2026-09-22
Test-coverage in-loop embedding calls (judge and suggestion cache keys, dedup) are never recorded


mgcronin:
The judge and the suggestion generator each call the embedder inside their loops, and none of that usage reaches the run's cost tracker:

- `judgeAmbiguous` (`server/src/lib/testcoverage/judge.ts`) embeds each batch's prompt for the semantic-cache key.
- `generateSuggestions` (`server/src/lib/testcoverage/suggestion-generator.ts`) embeds each cluster's prompt for the cache key, and then embeds every suggestion's text for dedup.

With a local embedder this costs nothing. With a cloud embedder (`bedrock`, `bedrock-sdk`, `openai`), it is real spend that the run budget never sees. It also never reaches `ai_token_usages`.

#58 (PR for #57/#58) records and prices the match phase's embeddings under the embedder that ran (`embed:<key>`). These in-loop calls are the rest of that usage.

Found while working on #58.

## Acceptance criteria
- [ ] Every embedder call made by a coverage run is recorded through `CoverageCostTracker` with `phase: "embedding"`, the embedder's key and the model it reported.
- [ ] A test with a priced cloud embedder shows these calls on the run budget.

PR #59, opened 2026-09-22
fix: Resolve #42, #43 — published Claude prices; test-coverage usage under what served it, unpriced fails closed

mgcronin:
Closes #42
Closes #43

Pricing follow-ups to PR #41 (#22). Both issues and the #41 re-review advisory touch the one price source, so they ship together.

## #42: every built-in Claude price is pinned to its published price

**Before:** `bedrock-gateway:us.anthropic.claude-fable-5` was billed at Haiku 4.5's $1/$5. The tests only checked that a rate was non-zero, so nothing caught it.

**Sources, read 2026-09-22:**
- Anthropic's [pricing page](https://platform.claude.com/docs/en/about-claude/pricing), for the `anthropic` provider.
- The AWS Price List API, offer `AmazonBedrockFoundationModels`, us-east-1, publication 2026-09-11, for `bedrock-gateway`. This is the machine-readable form of https://aws.amazon.com/bedrock/pricing/. That page renders its Anthropic table client-side, and WebFetch could not read it.

**What the sources say:** Bedrock has two on-demand SKUs per Claude 4.5+ model:
- *Global*, which matches Anthropic's price;
- *Regional* ("Regional CRIS"), at 1.1×.

A `us.anthropic.…` geo profile bills at the Regional SKU. Anthropic's page says the same thing: "Regional and multi-region endpoints include a 10% premium over global endpoints", from Sonnet 4.5, Haiku 4.5 and Opus 4.5 on.

**Changed rows** (USD per MTok: input / output / cache read / 5-minute cache write):

| Row | Was | Now |
|---|---|---|
| `bedrock-gateway:us.anthropic.claude-fable-5` | 1 / 5 (Haiku) | **11 / 55 / 1.10 / 13.75** |
| `bedrock-gateway:us.anthropic.claude-sonnet-5` | 3 / 15 | 2.20 / 11 / 0.22 / 2.75 |
| `bedrock-gateway:us.anthropic.claude-sonnet-4-6`, `anthropic.claude-sonnet-4-6` | 3 / 15 | 3.30 / 16.50 / 0.33 / 4.125 |
| `bedrock-gateway:us.anthropic.claude-opus-4-8` | 5 / 25 | 5.50 / 27.50 / 0.55 / 6.875 |
| `anthropic:claude-3-5-haiku`, `bedrock-gateway:anthropic.claude-3-5-haiku-20241022-v1:0` | 1 / 5 | 0.80 / 4 / 0.08 / 1 |

[...]

Issue #43, opened 2026-09-22
Test-coverage cost tracker records every call as bedrock-gateway and cannot price unpriced usage safely


mgcronin:
`server/src/lib/testcoverage/cost-tracker.ts` (`CoverageCostTracker.record`) records every judge or suggestion call under the provider `bedrock-gateway`, whatever provider actually served it. It also prices the call with `estimateCostUsd(modelId, …)` and passes no provider.

On a deployment where `AI_PROVIDER=anthropic` and `ANTHROPIC_BASE_URL` points at DeepSeek, this means:
- `ai_token_usages` rows for test-coverage runs name the wrong provider;
- the per-run budget estimate prices `claude-haiku-4-5` at Anthropic list rates, although DeepSeek serves and bills that name as `deepseek-flash` (https://api-docs.deepseek.com/guides/anthropic_api).

Passing the real provider alone is **not** a fix. On that deployment the model is unpriced (#22), `estimateCostUsd` returns `null`, and the `?? 0` in `record()` would make the per-run budget (`exceeded()` / `canAfford()`) unable to trip. That is the same failure mode PR #41 fixed for the autopilot ceiling. Unpriced usage has to fail closed here too.

`CoverageCostTracker` is built in `coverage-service.ts` from `runId`, `projectId` and `userId` only, so the provider first has to be threaded through from where the judge's `callModel` is wired.

Found in review of PR #41.

## Acceptance criteria
- [ ] Test-coverage usage is recorded under the provider that served it.
- [ ] The per-run budget treats unpriced usage as "cannot evaluate" (fail closed, or a clear refusal), never as $0. A test shows an all-unpriced run cannot overspend.

PR #74, opened 2026-09-22
fix: Resolve #57, #58 — suggestion budget checked per cluster; embeddings priced as the embedder that ran

### LINKED ISSUE #57: Test-coverage suggestion phase has no budget check between clusters
`generateSuggestions` (`server/src/lib/testcoverage/suggestion-generator.ts`) calls the model once per requirement cluster and returns the summed tokens. `runCoverageScoring` (`coverage-service.ts`) records that total through `CoverageCostTracker` only after every cluster has run, and checks `cost.exceeded()` only before the phase starts.

The judge phase is different: it records each batch and checks `exceeded()` between batches (#883). The suggestion phase has no such check, so once it starts it runs every cluster, whatever the budget:

- a priced run can go over `TESTCOVERAGE_BUDGET_CENTS` by the whole suggestion phase;
- an all-unpriced run (#43), where the judge phase had no ambiguous pairs, makes every suggestion call before the budget sees the first unpriced one. PR for #42/#43 makes the budget fail closed on unpriced usage, but it can only stop LLM work at the points where it is checked.

Found while working on #43.

## Acceptance criteria
- [ ] The suggestion phase records each cluster's usage as it happens and checks the budget before each cluster, the same way the judge does between batches.
- [ ] A test shows a suggestion phase stopping part-way when the budget is reached, and after the first unpriced call.

### LINKED ISSUE #58: Test-coverage embedding usage is priced at Claude Haiku's rate
`runCoverageScoring` (`server/src/lib/testcoverage/coverage-service.ts`) records the match phase's embedding usage as `cost.record({ phase: "embedding", embeddingTokens })`. It passes no model, so `CoverageCostTracker.record` (`cost-tracker.ts`) records it under provider `offline-stub` and model `HAIKU_MODEL_ID`. `resolveRate("offline-stub", HAIKU_MODEL_ID)` finds no exact row and then matches the Claude Haiku family, so every embedding token counts against the run budget at Bedrock's Haiku 4.5 input price.

That is another model's price. The embedder (`server/src/lib/rag/embedder.ts`) can be local (transformers.js, which costs nothing per token) or a cloud backend (`bedrock`, `bedrock-sdk`, `openai`), which has its own price. The token counts are estimates from `estimateTokens`, not what the embedder reports.

Found while working on #43. The PR for #42/#43 does not change this; embedding usage is priced exactly as before.

## Acceptance criteria
- [ ] Embedding usage is recorded under the embedder that ran (`getEmbedder().key` / `.model`), and priced through `resolveRate` like any other usage: $0 for a local backend, the published price for a cloud one, and unpriced where there is none.
- [ ] A test pins what a local-embedder run's embedding phase adds to the run budget.

mgcronin:
Closes #57
Closes #58

Two follow-ups to PR #59, both on the test-coverage run budget.

## #57: the suggestion phase checks the budget before each cluster

**Before:** `runCoverageScoring` checked `cost.exceeded()` once, before `generateSuggestions` started. It recorded the phase's total only after every cluster had run. So a run with a model that has no price, whose judge had nothing to judge, made every suggestion call before the budget saw the first unpriced token.

**Now:**
- `generateSuggestions` takes an optional `cost: SuggestionBudgetGuard`. `CoverageCostTracker` satisfies it.
- Every model call is recorded through the guard as it happens, under the provider and model that served *that* call. This also fixes the #59 review's low finding: a mid-phase provider fallback no longer prices every call at the last call's rate.
- `exceeded()` is checked before each cluster, the same way the judge checks between batches. A priced run stops part-way once the cap is reached. An unpriced run stops after the first call.
- `GenerateResult.budgetExceeded` reports the stop. The service maps it onto the run's `budgetExceeded` and the `suggest` done event.
- The service no longer records the phase total afterwards. The generator records it, so nothing is counted twice; a test pins `suggestionTokens` at exactly one call's 150.

## #58: embedding usage is priced at the embedder that ran

**Before:** `cost.record({ phase: "embedding", embeddingTokens })` passed no model. The tracker recorded it on `offline-stub` under `HAIKU_MODEL_ID`, and `resolveRate` family-matched that to Bedrock Haiku 4.5. About 12.5k tokens of local-embedder usage cost the run 2 cents of a 20-cent budget.

**Now:**
- The embedding record requires `embedder` (`getEmbedder().key`, read after `embed()`, because a failed backend may have been swapped for the hash stub) and `modelId` (the model `embed()` reported).
- The usage is recorded and priced under provider `embed:<key>`, through `resolveRate`, which is the one price source. `embed:` is a namespace of its own: the `openai` embedder must never pick up the `openai` chat provider's rows. A test pins this. `UsageProvider = ProviderKey | \`embed:${string}\`` widens `TokenEvent.provider` for this.
- Local backends (`offline`, `xenova`, `embeddinggemma`, `sidecar`) have `embed:<key>:default` rows at a genuine $0.
- Cloud rows use published prices only (USD per MTok, input only):

| Row | Price | Source (read 2026-09-22) |
|---|---|---|
| `embed:bedrock:amazon.titan-embed-text-v2:0` | 0.02 | AWS Price List API, offer `AmazonBedrock`, us-east-1, publication 2026-09-17, SKU `USE1-TitanEmbeddingV2-Text-input-tokens`: "$0.00002 per 1K tokens" |
| `embed:bedrock-sdk:amazon.titan-embed-text-v2:0` | 0.02 | same |
| `embed:openai:text-embedding-3-small` | 0.02 | https://developers.openai.com/api/docs/pricing (Standard) |
| `embed:openai:text-embedding-3-large` | 0.13 | same |
| `embed:openai:text-embedding-ada-002` | 0.10 | same |

- Anything else is **unpriced**: never another model's price. That includes Cohere on Bedrock (its Marketplace SKUs don't map cleanly to a model id) and Azure deployment names. Unpriced usage fails the budget closed, as #43 does for the LLM phases. An administrator can price such a model with `MODEL_PRICES` key `embed:<backend>:<model>`. USER_GUIDE says so.
- `published-embedding-prices.test.ts` pins every `embed:` row and checks that set, so adding or removing a row without a pinned price fails.

## Tests that fail before the fix

On the unfixed code, the new and updated tests fail: 26 in 5 files. Each fails for the reason it names. For example, the coverage-service local-embedder test fails with `expected 2 to be +0`: 2 cents at the Haiku rate, where the local embedder should add $0.

## Mutations (each applied alone, then restored; suites: testcoverage cost/suggestion/service/snapshot/judge + finops price tests, 139 tests)

| # | Mutation | Went red |
|---|---|---|
| M1 | Generator skips the per-cluster `exceeded()` check | 5: priced part-way stop, unpriced stop, already-exhausted, and both service #57 tests |
| M2 | Generator never records calls through the guard | 6, including per-call provider recording and #43 service test |
| M3 | Service does not pass `cost` to `generateSuggestions` | 3 |
| M4 | Service ignores `sugResult.budgetExceeded` | 2 |
| M5 | Embeddings recorded on `offline-stub` again | 4 |
| M6 | `embeddingUsageProvider` drops the `embed:` namespace | 15, including "an embedder key never picks up an LLM provider's row" |
| M7 | Drop the `embed:xenova:default` $0 row | 9, including the row-set pin |
| M8 | Titan V2 priced 10x | 2 |
| M9 | Service records the configured model instead of the one `embed()` reported | 1 |
| M10 | An embedding record with no embedder priced as Haiku | 1 |

One survivor, outside this change: deleting `cost` from the service's `judgeAmbiguous(...)` call left all 139 green. Filed as #73.

## Coverage (touched source files, lines / branches, over the suites above)

| File | Before (origin/main) | After |
|---|---|---|
| `testcoverage/suggestion-generator.ts` | 94.53 / 80.00 | 95.52 / 81.08 |
| `testcoverage/cost-tracker.ts` | 100 / 91.30 | 100 / 92.30 |
| `testcoverage/coverage-service.ts` | 88.00 / 64.00 | 87.90 / 66.66 |
| `finops/provider-rates.ts` | 84.21 / 72.72 | 96.10 / 85.22 |
| All four | 91.25 lines | 93.93 lines |

The floor is 80% lines. `coverage-service.ts` lines dropped by 0.1 point: the code is reshaped, and nothing that was covered before is uncovered now. The changes in `ai/types.ts`, `ai/token-tracker.ts` and `ai/cache-verification.ts` widen types only.

## Gate (run locally)
- `pnpm lint`: exit 0. The company-identifier scan found 0 of 43 private terms, and I also grepped the staged diff against the list: 0 hits.
- `pnpm typecheck`: exit 0.
- `pnpm test`: **exit 1, but not from this change.** Two `scripts/` tests timed out at their 5 s limit: `verify-no-company-identifiers-runner` "passes the gate…" and `gate-cwd-anchoring` "identical output…". The machine was at load average ~37 with several agents running (`uptime`). The scanner alone took 4.86 s wall for 2.3 s of CPU. This PR does not touch `scripts/`. Because pnpm stops at the first failing package, I ran the rest directly:
  - server: 1117 files / 17563 tests passed;
  - ui: 359 / 3359;
  - shared: 36 / 698;
  - embeddings-svc: 147;
  - copilot-svc: 78.
  CI is the authority on the two timeouts.
- `node scripts/verify-changelog-fragment.mjs --base origin/main`: passes, with two fragments: `57-suggestion-budget-per-cluster.md` and `58-embedding-usage-real-price.md`.

## Security
- No dependency was added or bumped, so I ran no audit.
- OWASP pass over the diff: no database query shapes changed, no shell, and no user input reaches a price lookup. `MODEL_PRICES` keys stay admin-set and schema-validated. No secrets.

## Write/read check
- The embedding row's persisted `estimatedCostUsd` is read back through the mocked `aITokenUsage.create`. It shows `provider: "embed:xenova"`, the model `embed()` reported, and `0`.
- `readBudget` sums unpriced tokens from those same rows. An unpriced embedding row is therefore counted by the persisted view as well as the in-memory one.
- No real-database test was added. The FK integration test from the #59 review is still open.

## Deferred, filed
- #72: the judge's and the generator's in-loop embedder calls (cache keys, suggestion dedup) are never recorded. They cost nothing with a local embedder, but a cloud embedder's spend there is invisible.
- #73: no test covers the service passing its cost guard to the judge (the M3b survivor above).
- Not done here: from the #59 review, the synthetic `AISession` staying `active`, and `recordAndFlush` draining the global pending counter.


mgcronin (commented):
## Code Review Summary — PR #74

Recommendation: COMMENT (merge-ready once CI is green). 0 blocking findings, 3 advisory findings (1 medium, 2 low).

This is an automated self-review, posted as a comment. It is not a human sign-off. The reviewing identity (`mgcronin`) also opened this PR, so GitHub cannot record an approval or a change request from it. A human decides whether to merge.

### Requirements: PASS
- [x] #57: the suggestion phase records each cluster's usage as it happens and checks the budget before each cluster, like the judge. `generateSuggestions` records per call and checks `cost.exceeded()` before each bucket.
- [x] #57: tests show the phase stopping part-way once the budget is reached, and after the first unpriced call. There are generator-level and service-level tests for each, plus a priced control that makes more than one call.
- [x] #58: embedding usage is recorded under the embedder that ran (`embed:<key>`, with the model `embed()` reported) and priced via `resolveRate`: $0 for local backends, the published price for Titan V2 and OpenAI, unpriced otherwise.
- [x] #58: a test pins what a local-embedder run's embedding phase adds to the budget ($0, previously 2 cents), including the persisted row's `estimatedCostUsd`.
- No scope creep and no unrelated reformatting. The pre-commit hook ran prettier on staged files only.

### Security: CLEAN
No new I/O, queries, or shell. Price keys come from server-side registry keys, and `MODEL_PRICES` stays admin-set and schema-validated. No dependency was added or bumped.

### Code quality: GOOD
The `embed:` namespace, and the `UsageProvider` type that widens `TokenEvent.provider`, keep embedder keys off LLM rows. One test pins that (`openai:gpt-4o` is priced; `embed:openai:gpt-4o` is not). The widened `ProviderKey` in `ai/` is type-only.

### Performance: CLEAN
One extra `exceeded()` call per cluster and one `record` per call. That is negligible.

### Tests: SUFFICIENT
- Falsifiability: 10 mutations are listed, and all went red. The reviewer re-read M3, which initially matched the *judge* call. The corrected mutation targets the suggestion call. The judge-wiring survivor it exposed is filed as #73.
- Read-back: the #58 service test reads the persisted `ai_token_usages` row through the Prisma mock, not the tracker's own tally.
- Gap, filed: #72 (in-loop embedder calls are not recorded).

### Documentation: UP TO DATE
Two changelog fragments (`57-…`, `58-…`), and the validator passes. USER_GUIDE's Budget guard section documents the per-cluster check and the `embed:<backend>:<model>` pricing key. ARCHITECTURE's cost-tracker bullets are corrected, including the stale 150¢ default, which is actually 20¢ (`DEFAULT_BUDGET_CENTS`).

### Scope of this review
I read the diff and the surrounding code of all 16 changed files. I checked the prices against the AWS Price List API (`AmazonBedrock` us-east-1, publication 2026-09-17) and OpenAI's pricing page. Locally, lint, typecheck, and the server, ui, shared, and svc suites passed. `pnpm test` exited 1 on two `scripts/` scanner timeouts under machine load, as described in the PR body. CI is pending at the time of posting. I did not exercise the running application.

server/src/lib/finops/provider-rates.ts
```diff
@@ -215,8 +217,35 @@ const RATES: ReadonlyMap<string, TokenRate> = new Map([
...
+  // #58 — embedders, under `embed:<registry key>` (see embeddingUsageProvider).
+  // An embedding call bills input tokens only. Sources (read 2026-09-22):
+  //   Titan Text Embeddings V2 — AWS Price List API, offer AmazonBedrock,
+  //     us-east-1, SKU USE1-TitanEmbeddingV2-Text-input-tokens: $0.00002 / 1K.
+  //   OpenAI — https://developers.openai.com/api/docs/pricing (Standard):
+  //     text-embedding-3-small $0.02, -3-large $0.13, ada-002 $0.10 per MTok.
+  // Any other cloud embedding model is UNPRICED until an administrator prices it.
+  ["embed:bedrock:amazon.titan-embed-text-v2:0", { inputPer1k: 0.002, outputPer1k: 0 }],
+  ["embed:bedrock-sdk:amazon.titan-embed-text-v2:0", { inputPer1k: 0.002, outputPer1k: 0 }],
+  ["embed:openai:text-embedding-3-small", { inputPer1k: 0.002, outputPer1k: 0 }],
```
mgcronin, line 229: **Medium, advisory: a behaviour change for self-hosted OpenAI-compatible embedders.** The `openai` backend also serves *OpenAI-compatible* endpoints (`EMBEDDINGS_OPENAI_BASE_URL`), for example a self-hosted vLLM or Ollama running `nomic-embed-text`. That model has no row, so its usage is now unpriced. Under #43's fail-closed rule, every test-coverage run on such a deployment skips judge and suggestion calls (`budget_exceeded`) until an administrator adds `"embed:openai:nomic-embed-text": {"inputPerMTok": 0, "outputPerMTok": 0}` to `MODEL_PRICES`. This is what #58 asks for ("unpriced where there is none"), and USER_GUIDE documents the key. Still, operators on such a deployment will notice it, and a release note is worth considering. The reverse risk is small: a paid compatible endpoint would have to use one of OpenAI's exact model ids to be priced at OpenAI's rate.

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -155,10 +170,12 @@ export class CoverageCostTracker {
...
     const usage = { promptTokens: prompt, completionTokens: completion, totalTokens: total };
     // #43 — the same price lookup the persisted `ai_token_usages` row uses.
-    // Without a provider there is no telling whose price applies: unpriced.
-    const usd = provider ? estimateUsageCostUsd(modelId, usage, provider) : null;
+    // Without a provider or model there is no telling whose price applies:
+    // unpriced, and not persisted (a usage row needs both).
+    const usd = provider && modelId ? estimateUsageCostUsd(modelId, usage, provider) : null;
     if (usd === null) this.unpricedTokens += total;
     else this.estimatedUsd += usd;
+    if (!provider || !modelId) return;
```
mgcronin, line 178: Low: a record with no provider or model is now counted as unpriced but **not persisted**. Before this PR, it was persisted with `provider: undefined`, and that row could not have been written correctly anyway. The consequence is that `readBudget`'s unpriced total, which is read from `ai_token_usages`, would not include such tokens, while the in-memory `view()` would. The types now require both fields, so only a cast can reach this path. Acceptable, and noted here so it is not a surprise.

server/src/lib/testcoverage/suggestion-generator.ts
```diff
@@ -259,6 +306,13 @@ export async function generateSuggestions(input: GenerateInput): Promise<Generat
       promptTokens += out.promptTokens;
       completionTokens += out.completionTokens;
       servedBy = { provider: out.provider, model: out.model };
+      input.cost?.record({
```
mgcronin, line 309: This records per call, under the provider and model that served *that* call. It resolves the #59 review's low finding for the suggestion phase: a mid-phase provider fallback no longer prices every call at the last call's rate. The judge's first-call/retry case within one batch (`judge.ts`) is unchanged.

server/src/lib/testcoverage/coverage-service.ts
```diff
@@ -139,7 +139,14 @@ export async function runCoverageScoring(
   emit({ phase: "match", state: "running" });
   const reqTexts = reqRows.map((r) => `${r.title}\n${r.body}`);
   const reqEmb = await embedder.embed(reqTexts);
-  cost.record({ phase: "embedding", embeddingTokens: estimateTokens(reqTexts) });
+  // #58 — under the embedder that ran (read AFTER embed: a failed backend may
+  // have been swapped for the hash stub) and the model it reports.
+  cost.record({
+    phase: "embedding",
+    embedder: embedder.key,
```
mgcronin, line 146: Low: `embedder.key` is read after `embed()`, so a hash-stub fallback is recorded as `embed:offline` ($0), not as the configured backend. `modelId` comes from the result, and M9 pins that. The key-after-embed ordering itself is not pinned by a test; the mock's key is static.

server/src/lib/testcoverage/coverage-service.ts
```diff
@@ -151,7 +158,12 @@ export async function runCoverageScoring(
   if (caseRows.length > 0) {
     const caseTexts = caseRows.map((c) => caseText(hydrateCase(c)));
     const caseEmb = await embedder.embed(caseTexts);
-    cost.record({ phase: "embedding", embeddingTokens: estimateTokens(caseTexts) });
+    cost.record({
+      phase: "embedding",
+      embedder: embedder.key,
+      modelId: caseEmb.model,
```
mgcronin, line 164: **Low — mutation survivor I found (measured, not inferred).** I deleted this entire `cost.record({ phase: "embedding", … caseEmb … })` call and re-ran `tests/lib/testcoverage` + `src/lib/finops`: **432 passed / 432, all green**.

The requirement-side record at line 146 is well covered (the `#58` service test pins the persisted row), but nothing covers this one, because that test's fixture has no test cases — which is why it can assert `expect(embedding).toHaveLength(1)`. In a real run the test-case corpus is usually the *larger* of the two, so on a cloud embedder its silent loss would under-charge the run budget by most of the embedding spend, and `exceeded()` would not fail closed when it should.

Suggestion: extend the `#58` service test (or add one) with a fixture that has `testCases`, and assert **two** `testcoverage.embedding` rows and the summed `embeddingTokens`. Cheap, and it makes the second arm of this change falsifiable. Non-blocking.

server/src/lib/finops/provider-rates.ts
```diff
@@ -215,8 +217,35 @@ const RATES: ReadonlyMap<string, TokenRate> = new Map([
...
   ["offline-stub:offline-stub", DEFAULT_RATE],
   ["offline-stub:default", DEFAULT_RATE],
+  // #58 — embedders, under `embed:<registry key>` (see embeddingUsageProvider).
+  // An embedding call bills input tokens only. Sources (read 2026-09-22):
+  //   Titan Text Embeddings V2 — AWS Price List API, offer AmazonBedrock,
+  //     us-east-1, SKU USE1-TitanEmbeddingV2-Text-input-tokens: $0.00002 / 1K.
+  //   OpenAI — https://developers.openai.com/api/docs/pricing (Standard):
+  //     text-embedding-3-small $0.02, -3-large $0.13, ada-002 $0.10 per MTok.
+  // Any other cloud embedding model is UNPRICED until an administrator prices it.
+  ["embed:bedrock:amazon.titan-embed-text-v2:0", { inputPer1k: 0.002, outputPer1k: 0 }],
```
mgcronin, line 227: **Low, advisory — agreeing with your own note at line 229, and sharpening it.** The priced set is a handful of exact model ids, while `bedrock`, `bedrock-sdk` and `openai` each accept any model id (Titan V1, Cohere on Bedrock, an Azure deployment name, a self-hosted OpenAI-compatible endpoint serving `nomic-embed-text`).

The part worth spelling out: the embedding record is the **first** usage a run makes (match phase), so on such a deployment `exceeded()` is true before the judge phase starts and **every** run does zero LLM work — not the partial stop that unpriced LLM discovery produces mid-judge. That is what #58 and #43 ask for, and `USER_GUIDE.md` documents the `MODEL_PRICES` key, so I am not treating it as blocking.

Worth considering as a follow-up: when an embedding record comes back unpriced, log once at `warn` with the exact `MODEL_PRICES` key to add (`embed:<backend>:<model>`). An operator currently has to work out from a `budget_exceeded` run that the *embedder*, not the judge model, was the unpriced one.

server/src/lib/testcoverage/suggestion-generator.ts
```diff
@@ -99,6 +127,11 @@ export interface GenerateResult {
    * none was made (every cluster a cache hit). Usage is recorded under these.
    */
   servedBy: { provider: ProviderKey; model: string } | null;
+  /**
+   * #57 — true when the cost guard stopped the phase before a cluster, so some
+   * clusters got no suggestions.
+   */
+  budgetExceeded: boolean;
```
mgcronin, line 134: nit (doc drift): with `budgetExceeded` added here, the `servedBy` comment three lines up still says "Usage is recorded under these." That is no longer true whenever a guard is wired — and the only production caller (`coverage-service.ts:336`) now always passes `cost`, so usage is recorded per call inside the loop and `servedBy` has **no remaining production consumer**. Either re-word the comment to say it is the last call's provider, kept for the guard-less path and callers' telemetry, or drop the field. Low confidence that it matters today; flagged so a future reader does not trust a stale contract.

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -188,7 +205,7 @@ export class CoverageCostTracker {
     void task.finally(() => this.inflight.delete(task));
   }
 
-  private ensureSession(provider: ProviderKey, model: string): Promise<boolean> {
+  private ensureSession(provider: UsageProvider, model: string): Promise<boolean> {
```
mgcronin, line 208: Informational: because the match-phase embedding is the first usage recorded, `ensureSession` now creates the run's synthetic `AISession` with `provider: "embed:xenova"` (previously `offline-stub`). I checked the read paths and nothing breaks — `ai_sessions.provider` is a plain `String` with no enum or zod gate, and `usage/usage-service.ts` aggregates the column as an opaque string — so the only visible effect is a new `embed:*` provider bucket in the admin usage views and CSV export, which is arguably an improvement.

Two small tidy-ups if you touch this again: `schema.prisma`'s doc comment on `AISession.provider` still enumerates only LLM keys (`copilot|bedrock-gateway|offline-stub|openai|azure|anthropic`), and the same for `conventionForProvider`'s "every other key (bedrock gateway, openai, azure, local, copilot)" now that an `embed:*` key reaches it.

mgcronin (commented):
## Code Review Summary — PR #74 (head b750b86)

Recommendation: merge with follow-ups — 0 blocking findings, 4 non-blocking findings (1 low test-gap, 1 low advisory, 2 informational/nit).

This is automated analysis published as a comment, not a human sign-off. The reviewing identity (`mgcronin`) is also the PR author, so GitHub cannot record an independent approval or change request from it. A human owns the merge decision.

Reviewed against `CLAUDE.md` (build/test gate, 80% line floor on touched code, `.changes/unreleased/<issue>-<slug>.md` fragments, never `CHANGELOG.md`, no version bumps, CI reading rules). There is no `.claude/delivery.md`; nothing here rests on an inferred convention.

### Requirements: #57 PASS, #58 PASS
- [x] #57 AC1 — the suggestion phase records each cluster's usage as it happens and checks `exceeded()` before each cluster (`suggestion-generator.ts` loop head + per-call `cost.record`). The service no longer records the phase total, so nothing double-counts.
- [x] #57 AC2 — tests show a priced phase stopping part-way (1 of 3 clusters, 1¢ cap) and an unpriced phase stopping after the first call, at both the generator and the service level.
- [x] #58 AC1 — embedding usage is recorded under `embed:<getEmbedder().key>` with the model `embed()` reported, priced through `resolveRate` only: $0 for the four local backends, published prices for Titan V2 / OpenAI, `null` (unpriced, fails closed) otherwise.
- [x] #58 AC2 — `a local-embedder run's embedding phase adds $0 to the run budget` is pinned at both the tracker and the service level, and the persisted `ai_token_usages` row is read back (`provider: "embed:xenova"`, reported model, `estimatedCostUsd: 0`).
- No scope creep. The one drive-by (`ARCHITECTURE.md` 150¢ → 20¢) matches `DEFAULT_BUDGET_CENTS`. No version bump; no `CHANGELOG.md` edit.

### Prices: VERIFIED INDEPENDENTLY (not taken on the author's word)
- Titan Text Embeddings V2: I downloaded the AWS Price List API offer `AmazonBedrock`, us-east-1. `publicationDate` **2026-09-17T23:16:09Z**; SKU `EDNQYEHQ3XQ8N7EY`, usagetype `USE1-TitanEmbeddingV2-Text-input-tokens` = **$0.00002 per 1K tokens** = $0.02/MTok. The table row `inputPer1k: 0.002` cents is exactly that. (Batch is $0.00001/1K — correctly not used.)
- OpenAI, from https://developers.openai.com/api/docs/pricing: `text-embedding-3-small` $0.02, `-3-large` $0.13, `ada-002` $0.10 per 1M input tokens. All three rows match, and `outputPer1k: 0` is right for an embedding call.
- The four zero rows (`offline`, `xenova`, `embeddinggemma`, `sidecar`) are the complete set of local backend keys in `rag/embedder.ts` / `embedder-registry.ts`; `bedrock`, `bedrock-sdk` and `openai` are real registered cloud backends whose defaults are exactly the priced model ids.

### Security: CLEAN (OWASP pass)
No new input surface. The `embed:` provider string is built from `getEmbedder().key`, which comes from `EMBED_BACKEND` config, never from a request. `MODEL_PRICES` stays admin-set and schema-validated. No query shapes changed, no scoping column lost, no shell, no secrets, no dependency added or bumped. `ai_token_usages.provider` and `ai_sessions.provider` are `String` columns with no enum or zod gate, so the widened `UsageProvider` cannot break a read path (`usage-service.ts` aggregates them as opaque strings).

### Tests: SUFFICIENT — claims checked, one survivor found
Run in a detached review worktree at the PR head (`pnpm install --frozen-lockfile --prod=false`, `pnpm --filter @metis/shared build`):
- **At head:** `tests/lib/testcoverage` + `src/lib/finops` = **432 passed / 432**, 40 files.
- **Tests fail before the fix — CONFIRMED.** I restored all 7 production files to the base (924556d) and kept the new tests: **26 failed / 406 passed** (4 files). The author's failure count matches exactly (they said 5 files; I measured 4 — the count of failing *tests* is identical).
- **My own mutation, M1 alone** (delete the per-cluster `exceeded()` block): **5 red**, matching the author's table.
- **My own mutation, a survivor the author did not list:** deleting the *test-case* embedding `cost.record(...)` in `coverage-service.ts` leaves **all 432 green**. Inline below.
- Coverage measured over those suites: `cost-tracker.ts` 100 lines / 92.30 branches, `suggestion-generator.ts` 95.52 / 81.08, `coverage-service.ts` 87.90 / 66.66, `provider-rates.ts` 97.40 / 96.59 — all above CLAUDE.md's 80% line floor.

### Gate (run by me, at the PR head)
- `pnpm typecheck`: **exit 0** (all 9 projects Done).
- `eslint` on the 12 changed `.ts` files: **exit 0**. I did not re-run whole-repo `pnpm lint`; CI's full gate is green.
- `node scripts/verify-changelog-fragment.mjs --base 924556d`: **passes** — 2 fragments contributed, one per issue.
- The author's local `pnpm test` exit 1 was two `scripts/` scanner tests hitting their 5 s limit under load — a timeout shape, not this diff (the PR touches no `scripts/`), and CI ran the full gate green.

### CI
Plain re-read of `gh pr checks 74`: **14 pass, 0 pending, 0 failing**, 2 `skipping` (`e2e`, `windows`, as normal). `api` pass and `api-outcome` pass, so no supersede ambiguity. `postgres-adapter` pass.

### Documentation: UP TO DATE
Two fragments (`57-…`, `58-…`), no `CHANGELOG.md` edit. `ARCHITECTURE.md` and `USER_GUIDE.md` both updated, including the `MODEL_PRICES` key form `embed:<backend>:<model>` that an operator needs when a cloud embedder is unpriced.

### Credit
The per-call recording under the serving provider quietly closes the #59 review's fallback finding; the `embed:` namespace with a row-set pin test (`published-embedding-prices.test.ts`) means a new row cannot be added without a published price; and both price sources check out to the cent against their primary sources, including the exact publication date cited.

### Scope of this review
Read both issues, PR #59's review, the full diff, and these files in full: `cost-tracker.ts`, `coverage-service.ts`, `suggestion-generator.ts`, `provider-rates.ts` (`resolveRate`/`overrideRate` order), `rag/embedder.ts` (backend keys and the hash-fallback swap), `rag/backends/index.ts`, the `AISession`/`AITokenUsage` schema, and `usage/usage-service.ts`. Not done: a live cloud-embedder run, and the running application.

Issue #75, opened 2026-09-22
Deploy follow-ups from #60/#51: prod RWO PVCs with 2 replicas, pgvector readiness, backup misses data dir, S3 smoke


mgcronin:
Deferred from the #60 / #51 bundle. Each item was found or confirmed while fixing those issues, and each is outside their scope.

- [ ] **Medium: `values-prod.yaml` mounts ReadWriteOnce uploads and LanceDB PVCs into 2+ server replicas.** `persistence.enabled: true` renders `<release>-server-uploads` and `<release>-server-lancedb` as RWO gp3 claims. Every server pod mounts them, and `server.replicaCount` is 2 with the server HPA's `minReplicas: 2`. A second replica scheduled on another node cannot attach them (Multi-Attach). Production keeps uploads on S3 and vectors in pgvector, so neither volume is needed there. The #60 bundle made the new `server-data` PVC opt-out for prod (`persistence.data.enabled: false`). The two older PVCs need the same treatment.
- [ ] **Medium: a Postgres without the `vector` extension boots cleanly under `VECTOR_STORE=pgvector` and only fails on the first vector write.** This was measured with the image smoke's Postgres arm against `postgres:16-alpine`. `/healthz` returned 200 and every probe passed except `pgvector` (`extension "vector" is not available`). A boot-time check, or a `/readyz` component, would fail fast instead of failing mid-ingest.
- [ ] **Low: the backup CronJob does not capture the server's data directory.** It mounts only the uploads and LanceDB PVCs. On SQLite the database is now `/app/server/data/metis.db` on the `server-data` PVC (#60), so `backup.enabled=true` backs up everything except the database.
- [ ] **Low: the S3 uploads backend (`UPLOAD_STORAGE_BACKEND=s3`) is not exercised by the image smoke.** It needs a bucket. A MinIO container on the smoke's private network would cover it, the same way `pgvector/pgvector:pg16` covers the database.
- [ ] **From #51, not done there:** the no-shell Windows spawn of the migration guard is still only tested with a faked platform (the Windows CI job is skipped, #2). The ~150 MiB Prisma CLI tree still ships in the runtime image; running `prisma migrate deploy` as a separate init step would drop it (`docs/OPERATIONS.md` > "Container Image Sizes").

PR #76, opened 2026-09-22
fix: Resolve #60, #51 — production values and Helm defaults start; smoke covers pgvector, Helm defaults, Dependabot

### LINKED ISSUE #51: Server image follow-ups from #46: Dependabot smoke, Prisma CLI path, Oracle checksum, smoke secrets
Non-blocking findings from the review of #46 (https://github.com/openzigs/metis/pull/46#pullrequestreview-5277032002), which made the server image start (#39). Findings 1 and 2 of that review are handled with #45.

- [ ] **Low — Dependabot PRs skip the boot smoke** (`.github/workflows/ci.yml`, smoke step `if:`). This matches the existing rule that images are not built for Dependabot PRs, but dependency bumps are the likeliest cause of the next image that won't start, and they would first fail on `main`. Consider building and smoking the server image for Dependabot PRs.
- [ ] **Low — the migration guard reaches into Prisma's internals.** `server/src/lib/db/migration-guard.ts` locates the CLI via `prisma/build/index.js`, which a Prisma upgrade could move. The error is clear and the smoke would catch it; resolve the CLI through a supported entry point if one exists.
- [ ] **Low — the no-shell Windows spawn is only tested with a faked platform**, and the Windows CI job is skipped (#2).
- [ ] **Low (pre-existing) — the Oracle Instant Client zip is downloaded without a checksum check** (`Dockerfile.server`, Oracle stage). Pin and verify a SHA-256.
- [ ] **Informational — the whole Prisma CLI (~150 MiB, including studio-core, pglite and react-dom) ships in the runtime image** so the migration guard can run `migrate deploy`. Running migrations as a separate init step would drop it (see `docs/OPERATIONS.md` > "Container Image Sizes").
- [ ] **Nit — `docs/OPERATIONS.md` still says "Node.js 20 runtime"** in the size table; the image now uses Node 22.
- [ ] **Nit — the smoke step passes its per-run random secrets on `docker run`'s command line**, so they are visible to `ps` on the runner while the step runs. Use `--env-file` with a temp file, or `-e NAME` with the value in the step's environment.

### LINKED ISSUE #60: Production deploy follow-ups from #55: pgvector smoke, Helm default can't start, compose vector store
Non-blocking findings from the review of #55 (https://github.com/openzigs/metis/pull/55#pullrequestreview-5278009658), which made the image start against Postgres (#45) and fixed the default data path (#54).

- [ ] **Medium — the Postgres smoke arm does not exercise the pgvector vector store.** Production (`deploy/helm/metis/values-prod.yaml`, `VECTOR_STORE=pgvector`) uses it, and it runs raw SQL through the relocated Postgres Prisma client. Add a pgvector probe to the Postgres arm of `scripts/lib/smoke-server-image.mjs` (the arm already runs `pgvector/pgvector:pg16`).
- [ ] **Medium (pre-existing) — the Helm chart's default cannot start.** With `DATABASE_URL` unset the server falls back to SQLite, and the chart also makes the root filesystem read-only, so `/app/server/dev.db` cannot be created at all (not merely "lost when the container is recreated", as #55's deferred list put it). Either require `DATABASE_URL` in the chart, or point the SQLite fallback at a writable mounted path, and add a smoke arm or chart test for the default values.
- [ ] **Low (pre-existing) — a compose production deployment uses the JSON vector store.** `docker-compose.prod.yml` inherits `VECTOR_STORE=${VECTOR_STORE:-local}` from the base file and mounts no volume for it. Decide the intended production default and set it explicitly.
- [ ] **Nit — `docs/OPERATIONS.md` "Container Image Sizes"**: the ui and embeddings rows are stale (CI measured 206.7 MB and 588.8 MB).
- [ ] **Agent memory** — persist the review's durable finding in `.claude/agent-memory/code-issue/`: "reverting the line that constructs the resolved Prisma client class fails no unit test; only ESLint's unused-variable rule and the image smoke catch it — top-level module wiring needs the smoke (or lint), not unit tests."

mgcronin:
Closes #60
Closes #51

Production deployment follow-ups from the reviews of #55 (#60) and #46 (#51). The new pgvector smoke arm found two more production defects, and this PR fixes them too: **the Helm production values could not start the server.**

Both are the "top-level wiring invisible to unit tests" shape named in #60's durable finding. The unit suite runs under `NODE_ENV=test`, where backend registration is skipped, so no unit test ever exercised the real order.

1. **`DISCUSSION_RATE_LIMIT_BACKEND=postgres` (values-prod) → exit at import.** `ai-rate-limit.ts`, `notify.ts` and `teams/inbound-rate-limit.ts` resolved their store at module scope. That runs before `createServer()` registers the Postgres factory: *"the Postgres store factory was not registered"*. **Fix:** each limiter now resolves its store on first use.
2. **`VECTOR_STORE=pgvector` (values-prod) → exit before listen.** `registerPgVectorStore()` ran after `createApp()`, and `createApp()` builds `documentsRouter`, which calls `getKnowledgeService()` → `getVectorStore()`. **Fix:** all five backend-factory registrations now run before `createApp()`.

Both were reproduced on an `origin/main` image with this branch's smoke. See "Evidence" below.

## #60

- **pgvector probe.** The Postgres arm now runs production's shared backends (`POSTGRES_ARM_BACKENDS`: `VECTOR_STORE=pgvector`, Postgres rate limit, SSO state and leader election). A `pgvector` probe then upserts, searches and counts a vector through the server's own `PgVectorStore`. The LanceDB probe is limited to the SQLite arms.
- **The Helm chart's default values can start.** I chose the writable-mount option, not "require DATABASE_URL":
  - The image sets `ENV DATABASE_URL=file:/app/server/data/metis.db`. The server's own default was `/app/server/dev.db` on the read-only root.
  - The chart mounts a writable volume at `/app/server/data`: a new `<release>-server-data` PVC, or an `emptyDir` when `persistence.data.enabled=false`, as `values-prod.yaml` sets.
  - It also mounts `/home/metis` (an `emptyDir`), because the server writes `~/.metis-sessions`.
  - Test: a new smoke arm, `helm-default`, boots the image read-only as uid 1001 with all capabilities dropped, no `DATABASE_URL`, and tmpfs only at the chart's mount paths. `render-tests.sh` Test 12 asserts the chart side. `image-build-wiring.test.mjs` fails if the smoke's path list and the chart's mounts drift apart.
  - **Behaviour change:** a SQLite deployment with no `DATABASE_URL` previously used `/app/server/dev.db`. The changelog says to copy that file or set `DATABASE_URL`.
- **`docker-compose.prod.yml`** sets `VECTOR_STORE: ${VECTOR_STORE:-lancedb}` and mounts `server_prod_data:/app/server/data`. Why LanceDB and not pgvector: compose runs one server container, and its `postgres:16-alpine` has no `vector` extension (measured, see below).
- **`docs/OPERATIONS.md`**: the ui size is now ~207 MB (Next.js 16) and embeddings ~589 MB, both from `api` run 35740221946.
- **Agent memory:** the #55 durable finding is persisted in `.claude/agent-memory/code-issue/project_module-wiring-needs-the-image-smoke.md`, extended with the two defects above. `pnpm agents:verify` passes.

## #51

- **Dependabot PRs** now run the disk-free, Buildx, `Build metis-server` and smoke steps. The ui, embeddings and sql-lineage builds and the size gate still skip Dependabot.
- **Prisma CLI:** it is now resolved through the package's declared `bin`, read via `prisma/package.json`, which is an explicit `exports` entry. It is no longer resolved as `prisma/build/index.js`. The change is in both the migration guard (`prismaCliFromManifest`) and the smoke's `migrations` probe.
- **Oracle Instant Client:** each architecture's zip is pinned by SHA-256, verified with `sha256sum -c`, and the build fails on a mismatch. Oracle's page no longer lists 23.9. I computed the SHA-256s from the served zips and checked them against the byte size and `cksum` Oracle published for 23.9, using an archived copy of the page:

  | Zip | Bytes | cksum |
  |---|---|---|
  | x64 | 72,900,819 | 423186373 |
  | arm64 | 65,159,765 | 1682780028 |

- **Smoke secrets:** `JWT_SECRET`, `VAULT_MASTER_KEY`, `EMBEDDINGS_TOKEN`, `DATABASE_URL` and `POSTGRES_PASSWORD` are passed as a bare `--env NAME`. Their values go in the docker CLI's environment (`spawnSync` `env`), never in its argv.
- **`docs/OPERATIONS.md`**: "Node.js 20 runtime" is now "Node.js 22 runtime", ~125 MiB (arm64 `du`).
- **Deferred to #75:** the Windows spawn item (#2) and the Prisma CLI init-step item, along with the other follow-ups found here.

## Evidence (local, arm64, Docker 29.5.3)

- **This branch's image:** `PASS: metis-server:smoke60 passes every arm: sqlite, postgres, helm-default`. Each arm passed 8 probes.
- **`origin/main` image, this branch's smoke:**
  - `--arm helm-default` → `unable to open database file: ./dev.db`
  - `--arm postgres` → `DISCUSSION_RATE_LIMIT_BACKEND=postgres selected but the Postgres store factory was not registered`
- **Intermediate build**, with the ordering fix but not the lazy limiters: the Postgres arm failed with `VECTOR_STORE=pgvector selected but the pgvector store factory was not registered`, in `documentsRouter` → `createApp`.
- **Postgres without pgvector** (`--postgres-image postgres:16-alpine`): the `pgvector` probe failed with `extension "vector" is not available`. The server itself booted, which is a gap tracked in #75.
- **Wrong Oracle SHA-256** (`docker build --target oracle-client`): `/tmp/ic.zip: FAILED`, and the build exits 1.

## Mutations (each restored afterwards)

| Mutation | Result |
|---|---|
| `migration-guard.ts` reverted to `origin/main` | 7 tests red |
| `prismaCliFromManifest` ignores `bin` | 1 red |
| `server.ts` reverted (registration after `createApp`) | 5 red (`server-backend-registration-order`) |
| Three limiters reverted to module-scope resolution | 3 red (`rate-limit-store-lazy-resolution`) |
| Secrets back on argv as `--env K=v` | 5 red |
| `POSTGRES_PASSWORD` back on argv | 2 red |
| Postgres arm without production backends | 1 red |
| `helm-default` without hardening | 1 red |
| Probe `arms` filter ignored | 2 red |
| `/home/metis` dropped from `HELM_DEFAULT_WRITABLE_PATHS` (chart/smoke drift) | 1 red |
| `sha256sum -c` line removed | 1 red |
| SQLite `ENV` moved out of `data/` | 2 red |

`render-tests.sh` Test 12 and the ci.yml, compose and Oracle wiring tests were written first and were red before each change: 4 chart assertions, 6 wiring tests, and the Oracle ordering test.

## Coverage (touched code)

| File | Stmts | Branches | Funcs | Lines |
|---|---|---|---|---|
| `scripts/lib/smoke-server-image.mjs` | 100 | 97.59 | 100 | 100 |
| `migration-guard.ts` | 100 | 85.18 | 100 | 100 |
| `ai-rate-limit.ts` | 100 | 92.85 | 100 | 100 |
| `notify.ts` | 95 | 100 | 92.3 | 94.23 |
| `inbound-rate-limit.ts` | 100 | 93.33 | 100 | 100 |

Before this PR, the #46 review reported the smoke at 100 / 98 / 100 / 100 and the migration guard at 80.95% branches.

## Gates (local)

- `pnpm lint`: exit 0, including the private-terms scan.
- `pnpm typecheck`: exit 0.
- `pnpm test`: the fan-out stopped on one `scripts` failure: `gate-cwd-anchoring.test.mjs` timed out at 5000 ms. It is pre-existing: the same failure reproduces on a clean `origin/main` worktree, and the test passes alone (9/9).
- Other packages:
  - `@metis/server`: 1118 files / 17552 tests passed.
  - `@metis/ui`: 1 failure (`workbench-chat-render`). It passes alone (3/3), and this PR touches no `ui/` file.
  - shared, copilot-svc, embeddings-svc and ui-kit: all passed.
- `bash scripts/chart-validate.sh`: 162 assertions passed. `helm lint` passes on default, dev and prod values.
- `pnpm agents:verify` and `pnpm changelog:verify`: pass.
- `docker compose config` on the prod overlay renders `VECTOR_STORE: lancedb` and the `server_prod_data` volume.

## Security

- No dependency changes, so I did not run a dependency audit.
- `docker` is still called with argv arrays and no shell.
- Secrets have moved off argv. `DATABASE_URL` in the image `ENV` is a file path, not a credential.
- The Oracle supply-chain download is now integrity-checked.
- The pgvector probe uses fixed strings only.
- The chart adds only `emptyDir` and PVC mounts. It adds no privileges, and the root filesystem stays read-only.

Not verified: that `cache-to: type=gha` works under Dependabot's restricted token. I expect it to, because Dependabot runs can write the Actions cache for their own ref. It will first run on the next Dependabot PR.


.github/workflows/ci.yml
```diff
@@ -179,7 +180,6 @@ jobs:
       # `load: true` puts each image in the runner's daemon so the size gate
       # below can `docker image inspect` it. #801 — run-scoped tag.
       - name: Build metis-server
-        if: ${{ !startsWith(github.head_ref, 'dependabot/') }}
         uses: docker/build-push-action@53b7df96c91f9c12dcc8a07bcb9ccacbed38856a # v7.3.0
```
mgcronin, line 183: **Medium — the one new failure mode this step can hit under Dependabot.** Line 192 keeps `cache-to: type=gha,scope=metis-server,mode=max` with no `ignore-error=true`. Dependabot PRs run with a restricted token, and a refused Actions-cache *write* surfaces from BuildKit as a cache-export error that fails the whole build step, not a warning. If that happens, every Dependabot PR goes red on `api` — strictly worse than the skip this PR removes, and it fails on the export *after* a successful build, so the message will not look like a permissions problem.

The PR body already flags this as unverified ("first exercised on the next Dependabot PR"). The insurance is one token:

```yaml
cache-to: type=gha,scope=metis-server,mode=max,ignore-error=true
```

It costs nothing when the write succeeds, and `cache-from` keeps working either way. I would take it before merge rather than learn the answer from a red Dependabot PR. (Confidence: medium — I could not test a Dependabot-token run; the behaviour is well attested for restricted-token PR runs.)

scripts/lib/smoke-server-image.mjs
```diff
@@ -51,8 +66,40 @@ export const DEFAULT_TIMEOUT_S = 120;
...
+ * The configurations the image must boot under, in the order they run: SQLite with
+ * the image's own default, Postgres (+ pgvector) as production runs it, and the
+ * Helm chart's default values (#60).
+ */
+export const ARMS = Object.freeze(["sqlite", "postgres", "helm-default"]);
+/**
+ * Where the Helm chart's default values let the server container write: every
+ * `mountPath` the server Deployment mounts (deploy/helm/metis/templates/
+ * _components.tpl, with the paths from values.yaml `persistence.*`). Everything else
+ * is read-only (`readOnlyRootFilesystem: true`). `helm-template-wiring.test.mjs`
```
mgcronin, line 79: **Low — the pointer to the drift gate names a file that does not exist.** `helm-template-wiring.test.mjs` is not in `scripts/lib/` (only `corpus-restore-wiring.test.mjs` and `image-build-wiring.test.mjs` are); the test that keeps this list equal to the chart lives in `scripts/lib/image-build-wiring.test.mjs`, which `render-tests.sh:591` names correctly.

This is the one comment whose whole job is to tell the next maintainer where the guard is, so the wrong name costs more than its size. Worth fixing to `image-build-wiring.test.mjs`.

Credit where due: I mutated the gate in the direction your table does not cover — I *added* a `/app/server/logs` mount to `_components.tpl` — and it went red too. The equality check catches drift both ways.

docs/OPERATIONS.md
```diff
@@ -322,9 +322,9 @@ The release pipeline ships three images:
 
 | Image | Built from | Current size | Budget |
 |---|---|---|---|
-| `metis-ui` | `Dockerfile.ui` (Next.js 15 standalone, alpine) | **~197 MB** | ≤ 350 MB ✅ |
+| `metis-ui` | `Dockerfile.ui` (Next.js 16 standalone, alpine) | **~207 MB** (amd64, `api` run 35740221946) | ≤ 350 MB ✅ |
 | `metis-server` | `Dockerfile.server` (`node:22-trixie-slim`, glibc, prod-only deps, slimmed) | **~1,045 MB** (amd64, measured #45) | ≤ 1,170 MB — its own budget, see below |
-| `metis-embeddings` | `Dockerfile.embeddings` (bookworm-slim, glibc) | **~423 MB** | exempt (sidecar) |
+| `metis-embeddings` | `Dockerfile.embeddings` (bookworm-slim, glibc) | **~589 MB** (amd64, `api` run 35740221946, built with `BAKE_MODELS=0`) | exempt (sidecar) |
```
mgcronin, line 327: **Low — the same number lives in a second place, and only this one moved.** `docs/ARCHITECTURE.md:145` still renders the embeddings sidecar as `~423 MB image` in its diagram box, which this row now contradicts (~589 MB). That is the "budget rationale lives in more than one file" trap this repo has hit before; a grep for the old figure when correcting one site is the cheap defence.

The figures here are right — I pulled them from the cited run myself: `metis-ui:ci-35740221946 207.1 / 350` and `metis-embeddings:ci-35740221946 588.8 exempt`. Good practice naming the run id in the table; it made this verifiable in one command.

docs/OPERATIONS.md
```diff
@@ -396,7 +396,7 @@ What is left, measured in the amd64 image (MiB, from `du`):
 | Component | Size | Notes |
 |---|---|---|
 | Oracle Instant Client (basiclite 23.9, `.so` only) | ~138 | Thick mode (pre-12c password verifiers); did not load in this image — loads since #39 |
-| Node.js 20 runtime (`/usr/local`) | ~104 | The `node` binary alone is 97 |
+| Node.js 22 runtime (`/usr/local`) | ~125 | The `node` binary alone is 117 (arm64, `du`, #51). #34 measured ~104 / 97 on the Node 20 base the image no longer uses |
```
mgcronin, line 399: nit — this table's caption says "measured in the amd64 image (MiB, from `du`)", and this row is now an **arm64** measurement. The row is honest about it inline, which is better than a silent swap, but a reader scanning the column will compare an arm64 number against amd64 neighbours (Oracle ~138, LanceDB ~99 …). Either re-measure on amd64 for consistency, or add "(arm64 where noted)" to the caption.

The substantive fix — Node 20 → Node 22 — is correct: the runtime base is `node:22-trixie-slim`.

server/src/lib/discussions/ai-rate-limit.ts
```diff
@@ -66,8 +66,14 @@ export function loadThreadAIRateLimitConfig(
...
  * window persists across calls within a process.
+ *
+ * #60 — resolved on FIRST USE, not at import. A shared backend's factory
+ * (`DISCUSSION_RATE_LIMIT_BACKEND=postgres`, production's setting) is registered by
+ * `createServer()`, which runs after every module is imported; resolving here made
+ * the server exit at import with "the Postgres store factory was not registered".
  */
-let store: RateLimitStore = resolveRateLimitStore();
+let store: RateLimitStore | undefined;
+const currentStore = (): RateLimitStore => (store ??= resolveRateLimitStore());
```
mgcronin, line 76: **Low — the fix is right, but it trades fail-fast for fail-late.** Deferring resolution to first use is the correct answer to the import-order bug (I verified it: reverting all three limiters turns `rate-limit-store-lazy-resolution` red with the exact production error). The side effect is that a genuinely broken shared backend — a factory that never registers on some entry path, a typo'd backend name — used to kill the process at import and now surfaces as a 500 on the first rate-limited request, in production, under load.

Since `createServer()` now registers the factories *before* anything resolves them, a one-line warm-up right after that block would restore the loud boot failure without reintroducing the ordering bug:

```ts
resolveRateLimitStore(); // fail loud at boot, not on the first discussion reply
```

Separately, `__setThreadAIRateLimitStore` (and the notify twin) now calls `currentStore()` purely to return `prev`, so injecting a stub *resolves the real store as a side effect* — under a selected-but-unregistered backend that throws where it previously did not. `store` (possibly `undefined`) as the returned previous value would avoid it. Both are non-blocking.

server/src/server.ts
```diff
@@ -174,6 +174,57 @@ export interface CreateServerOptions extends CreateAppOptions {
...
 export function createServer(opts: CreateServerOptions = {}): MetisServer {
+  // #60 — the shared-backend factories are registered BEFORE `createApp()`, because
+  // building the routers resolves some of these backends: `documentsRouter` builds
+  // the KnowledgeService, whose constructor calls `getVectorStore()`. Registered
+  // after it (as they were), `VECTOR_STORE=pgvector` — the production Helm setting —
+  // threw "the pgvector store factory was not registered" and the server exited
+  // before listening. No unit test could see it (they run under NODE_ENV=test,
+  // where this block is skipped); the image smoke's Postgres arm, which runs
+  // production's backends, did.
+  if (process.env.NODE_ENV !== "test") {
```
mgcronin, line 185: Good fix, and the right one: I confirmed the old block sat inside an identical `NODE_ENV !== "test"` guard further down, so this is a pure move with no change to *what* runs — only *when*. Reverting the file turns 5 tests red, as claimed.

One caveat worth knowing rather than acting on: `server-backend-registration-order.test.ts` asserts on the **source text** of this file (`^\s+registerX\(\);$` before `const app = createApp(opts);`). That is a defensible choice — `src/server.ts` is excluded from coverage and the block is skipped under `NODE_ENV=test`, so there is nothing else to assert on — but it means a *correct* refactor (extracting these five calls into `registerSharedBackends()` called before `createApp`) would fail the test. The test comment should probably say so, since the next person to touch this file will hit it.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking findings, 8 non-blocking (1 medium, 3 low, 4 nit/informational)

This is an automated analysis published as a comment, not a human sign-off. The reviewing identity (`mgcronin`) is also the PR author, so GitHub cannot record an approval or a change request from it. A human still owns the merge decision.

## Code Review Summary — PR #76 (head `b0dc2c8`)

### Requirements: PASS

**#60**
- [x] pgvector probe on the Postgres arm, with production's shared backends (`POSTGRES_ARM_BACKENDS`). CI log for run 35756748880 shows `OK: [postgres] probe pgvector` and `PASS: [postgres] … 8 probe(s)`.
- [x] The chart's default values can start: image `ENV DATABASE_URL=file:/app/server/data/metis.db`, a writable `data` volume (PVC or emptyDir) plus `/home/metis`, **with a test** — `render-tests.sh` Test 12 (162 assertions pass locally) *and* a third smoke arm (`OK: [helm-default] …` for all 8 probes in CI).
- [x] `docker-compose.prod.yml` pins `VECTOR_STORE: ${VECTOR_STORE:-lancedb}` on a named volume, with the reason (`postgres:16-alpine` ships no `vector`) recorded next to it.
- [x] OPERATIONS.md ui/embeddings rows. **Re-measured from the cited run:** `metis-ui:ci-35740221946 207.1 / 350`, `metis-embeddings … 588.8 exempt`, `metis-server 1045.5 / 1170`. The doc's ~207 / ~589 match. `next: ^16.3.3` in `ui/package.json` confirms the "Next.js 16" edit.
- [x] Durable finding persisted (`project_module-wiring-needs-the-image-smoke.md`); `pnpm agents:verify` passes here (index 931B, well inside the 17,500B budget).

**#51**
- [x] Dependabot PRs build and smoke `metis-server` only; ui/embeddings/sql-lineage builds and the size gate still carry the `dependabot/` skip, and a test asserts both directions.
- [x] Prisma CLI via the declared `bin`. **Checked in `server/node_modules/prisma/package.json` (7.8.0):** `"./package.json"` is an explicit `exports` entry and `bin.prisma = "build/index.js"`, so the new resolution is supported and behaviour-identical today.
- [x] Oracle Instant Client SHA-256. **Verified independently, not taken on trust:** I downloaded both zips and hashed them — x64 `664a88e5…079b3` and arm64 `6fd71c89…4dc16` match the pins byte-for-byte, and `Content-Length` matches the 72,900,819 / 65,159,765 the PR body cites.
- [x] Smoke secrets off docker's argv (`--env NAME`, values in the CLI's environment), including `POSTGRES_PASSWORD`.
- [x] "Node.js 20 runtime" → Node 22.
- Deferred items (Windows spawn test, Prisma-CLI-as-init-step) are filed in #75, which also carries the prod RWO-PVC and pgvector-readiness findings. Reasonable scoping.

### The two defects found along the way
Both are real and both are proven. The `helm-default` and `postgres` arms are the first tests in this repo that can see import-time and registration-order wiring, and they immediately caught that **the production Helm values could not boot**. That is the gate paying for itself on the first run.

### Security: CLEAN
- Supply chain **improved**: the Oracle download is now integrity-checked and the build fails on mismatch (I confirmed `sha256sum -c` sits between `curl` and `unzip`, `&&`-chained, and removing it turns the wiring test red).
- Secrets moved from argv to the docker client's environment — a real reduction in exposure on a shared runner.
- `docker` is still invoked with argument arrays, no shell. The pgvector probe interpolates only `process.pid`.
- The chart change adds only `emptyDir`/PVC mounts; `readOnlyRootFilesystem`, `runAsNonRoot`, dropped caps and `fsGroup: 1001` are untouched. I checked `fsGroup: 1001` explicitly (`_helpers.tpl:193`) — without it the new RWO `server-data` PVC would be root-owned and uid 1001 could not create `metis.db`, so this is load-bearing for the fix.
- The image `ENV DATABASE_URL` is a file path, not a credential. The base `docker-compose.yml` sets `DATABASE_URL` explicitly, so dev compose does not silently move its database.
- No dependency changes.

### Tests: SUFFICIENT — and falsifiable
I re-ran the author's mutations rather than accepting the table. Every one reproduced exactly:

| Mutation (mine, each restored) | Result |
|---|---|
| `server/src/server.ts` reverted to base | **5 red** (`server-backend-registration-order`) |
| three limiters reverted to module-scope resolution | **3 red**, with the real error `DISCUSSION_RATE_LIMIT_BACKEND=postgres selected but the Postgres store factory was not registered` |
| `migration-guard.ts` reverted to base | **7 red** |
| `sha256sum -c` line deleted | **1 red** |
| SQLite `ENV` moved out of `data/` | **2 red** |
| `/home/metis` dropped from `HELM_DEFAULT_WRITABLE_PATHS` | **1 red** |
| probe `arms` filter ignored | **2 red** |
| **new mount added to `_components.tpl`** (drift in the *other* direction) | **1 red** |

That last one is mine: the chart/smoke drift gate fires both ways, not just when the smoke list shrinks.

Coverage re-measured, not quoted: `scripts/lib/smoke-server-image.mjs` = **100 / 97.59 / 100 / 100**, matching the PR body exactly, against the project's 80% floor. Package thresholds are 80/80/80/80 in both `scripts` and `server` (`src/server.ts` is coverage-excluded, which is why the source-text ordering test exists).

### Gates run by me, in the PR worktree
- Touched tests: `scripts` 87 passed (2 files); `server` 24 passed (3 files).
- `eslint --max-warnings=0` on the 8 changed server files: clean.
- `bash deploy/helm/metis/tests/render-tests.sh`: **162 assertions passed**.
- `pnpm agents:verify`, `pnpm changelog:verify`: pass.
- `gh pr checks 76`: 16 pass, `e2e`/`windows` skipping, **0 pending**, all bound to head `b0dc2c88` (checked via `check-runs`). `api` and `api-outcome` are both green — not a supersede. CI's own `Test` step is the full `pnpm test` fan-out and it passed, which answers the two local flakes the PR body reports.
- I did not build the image locally; the three-arm CI smoke log is the evidence.

### Documentation: UP TO DATE
Changelog fragment `.changes/unreleased/60-production-deploy-can-start.md` is present, per-fragment as `CLAUDE.md` requires, and it names the SQLite migration step operators need. OPERATIONS.md, the chart README and NOTES.txt all move together. Two stale duplicates elsewhere are noted inline.

### Findings (all non-blocking; details inline)
1. **Medium** — `cache-to: type=gha` has no `ignore-error=true` on the step that now runs for Dependabot (ci.yml).
2. **Low** — `smoke-server-image.mjs:79` points at `helm-template-wiring.test.mjs`, which does not exist.
3. **Low** — `docs/ARCHITECTURE.md:145` still carries the old ~423 MB embeddings figure.
4. **Low** — the limiters now fail late rather than at boot; a boot-time resolve would restore fail-loud.
5. nit — an arm64 `du` figure inside a table captioned "measured in the amd64 image".
6. nit — `kind-smoke.sh` prints "render-tests.sh (109 assertions)"; it is 162.
7. nit — `runArm`'s option is still called `database` while its values are arms.
8. Informational — the `helm-default` arm models chart volumes as `tmpfs mode=1777`.

### Scope of this review
Read all 26 changed files, plus `_helpers.tpl`, `secret.yaml`, `pvc.yaml`, both compose files, `kind-smoke.sh`, `verify-image-size.mjs` (unchanged and still consistent), issues #60, #51, #75 and the reviews of #46 and #55. Ran the gates above and 8 mutations. Verified two external facts by fetching them: both Oracle SHA-256 pins, and the three image sizes from CI run 35740221946. Did not build the image, exercise a real Kubernetes rollout (`kind-smoke` deliberately does not wait for pod-Ready), or test the S3 uploads backend.

scripts/lib/smoke-server-image.mjs
```diff
@@ -214,22 +300,45 @@ export function buildRunArgs({
...
+          // deploy/helm/metis/templates/_helpers.tpl `metis.containerSecurityContext`.
+          "--read-only",
+          "--user",
+          HELM_RUN_AS,
+          "--cap-drop",
+          "ALL",
+          "--security-opt",
+          "no-new-privileges",
+          // Each chart volume (an emptyDir or a PVC) is writable by the pod's user.
+          ...HELM_DEFAULT_WRITABLE_PATHS.flatMap((p) => ["--tmpfs", `${p}:rw,exec,mode=1777`]),
```
mgcronin, line 335: Informational — worth recording what this arm can and cannot prove. `--tmpfs <p>:rw,exec,mode=1777` gives world-writable directories; the chart's emptyDir and PVC volumes are writable because the pod sets `fsGroup: 1001` (`_helpers.tpl:193`) — I checked that specifically, because without it the new RWO `server-data` PVC would mount root-owned `0755` and uid 1001 still could not create `metis.db`. So the model is faithful *today*, and the arm is genuinely load-bearing.

What it would not catch is a regression that depends on *owning* the directory rather than merely being able to write it (a `chown`, or a mode check), or a volume whose ownership `fsGroup` does not reach. If that is cheap to tighten, `--tmpfs <p>:rw,exec,mode=0770,uid=1001,gid=1001` models the cluster more exactly. Not blocking — the boot path is what this arm is for, and it works.

scripts/lib/smoke-server-image.mjs
```diff
@@ -432,7 +550,8 @@ async function runArm(io, o) {
     log(`OK: [${o.database}] /healthz returned 200`);
 
     const failed = [];
-    for (const probe of probes) {
+    const armProbes = probes.filter((p) => !p.arms || p.arms.includes(o.database));
```
mgcronin, line 553: nit — the rename from `--database` to `--arm` stopped at the public surface: `runArm`'s option is still `o.database`, so this filter reads `p.arms.includes(o.database)` and the logs interpolate `[${o.database}]` to print `[helm-default]`, which is not a database. Renaming the field to `arm` inside `runArm` and its `@param` block would finish the change; it is contained to one function.

deploy/helm/metis/tests/render-tests.sh
```diff
@@ -578,6 +579,46 @@ assert_contains "dev: embeddings still probes /readyz" "path: /readyz" "${DEV_EM
 assert_contains "embeddings uses the baked sidecar image" \
   "image: ghcr.io/openzigs/metis-embeddings-svc" "${EMB}"
 
+# ---------------------------------------------------------------------------
+echo "Test 12: the DEFAULT values can start — issue #60"
```
mgcronin, line 583: Test 12 is the right test for the requirement — it asserts the read-only root, a writable volume at each of the five paths, the PVC vs emptyDir split across default/no-persist/prod, the EFS toggle, *and* that `data` is listed before the volumes nested inside it. All 162 assertions pass locally.

nit, adjacent: `scripts/kind-smoke.sh:59` prints `ok "render-tests.sh (109 assertions)"`. It was already stale before this PR and is now further off (162). Since the count is hard-coded in a success message, it can only ever drift — either drop the number from the message or have the script echo the real total.

deploy/helm/metis/templates/_components.tpl
```diff
@@ -150,12 +150,19 @@ spec:
...
+            {{- /*
+              #60 — the root filesystem is read-only, so every path the server writes
+              needs a volume, or the chart's DEFAULT values cannot start: its home
+              (the image user's, ~/.metis-sessions) and its data directory (the SQLite
+              default, repo archive extracts). `data` is listed before the volumes
+              nested in it. scripts/lib/smoke-server-image.mjs (`helm-default` arm)
+              boots the image under exactly these writable paths.
+            */}}
+            - name: home
+              mountPath: /home/metis
```
mgcronin, line 163: Credit: this is a genuine simplification, not just an addition. The old template duplicated the uploads/lancedb mounts across an `if/else if` on `persistence.enabled`; the mounts are now unconditional for the server and only the *volumes* decide PVC vs emptyDir. I checked the `persistence.enabled=false` path still defines every referenced volume (`uploads`/`lancedb` emptyDir), so no dangling volume reference — and the render tests cover default, no-persist, prod and EFS.

The nested-mount ordering comment is worth keeping: `data` before `data/uploads` and `data/lancedb` is exactly the kind of constraint that is invisible until it breaks.

PR #80, opened 2026-09-22
fix: Resolve #67, #68 — no raw exception text in degraded-doc warnings; log Errors with message and stack

### LINKED ISSUE #67: Degraded generated docs still expose raw exception text in section warnings (follow-up to #52)

### LINKED ISSUE #68: Server logger writes an Error in log metadata as {} — message and stack are lost

mgcronin:
Closes #67
Closes #68

## #67 — degraded docs no longer surface raw exception text

`synthesizeFinalDocument` built a `section-failed` warning from `String(err)`
(`server/src/lib/docs-gen/holistic-synthesizer.ts`). `sectionFailedWarning`
appends up to 300 characters of that detail to a message which is persisted in
`GeneratedDocument.warnings`, returned by `GET /projects/:projectId/docs/:docId`
and rendered by `DegradedWarningsBanner` — so provider response bodies, absolute
server paths and SQL text reached the browser. That is the exposure #52 closed
for a *failed* row's `errorMessage`, reopened on the *degraded* path.

**Write path.** The section's error is mapped through the fixed
`generationFailureMessage` vocabulary before the warning is built. The raw error
still goes to the `log.warn` immediately above it, which is where it belongs.

**Read path.** `publicDocWarnings` (new, in `generation-failure-message.ts`,
beside `publicGenerationErrorMessage`) re-derives the detail of any persisted
`section-failed` warning that predates this change, on both `GET /:docId` and
`PATCH /:docId`. The legacy text is read only to *classify* it, never echoed, so
a stored provider 402 still comes back as a balance problem. Every other warning
kind is returned byte-identical, numeric `ratio`/`threshold`/`tier` included.

**Why the marker can't fail open.** `sectionFailedWarning` marks what it builds
`detailSafe: true`, and that is the read path's discriminator (no pre-#67 row can
carry it). On its own that would mean a *new* call site passing `String(err)`
gets stamped safe and sails through. `server/tests/docs-gen-warning-detail.enumeration.test.ts`
re-derives the call sites from the sources on disk and fails the build on one
that passes an exception's own text — including an arm that asserts the matcher
still matches the offending shapes, so deleting a pattern cannot leave it green.

The two other `sectionFailedWarning` call sites were audited and pass
METIS-authored text: the DB-schema synthesizer joins `ProseFailureReason` slugs
(a closed enum, never `f.detail`, which does hold `err.message`), and the
empty-section case passes a literal.

## #68 — an Error in log metadata is no longer `{}`

`redact()` rebuilt every metadata object from `Object.entries(...)`, and an
`Error`'s `name` / `message` / `stack` are not own *enumerable* properties, so
`log.error("…", { err })` was written as `"err":{}`.
`winston.format.errors({ stack: true })` does not cover this — it handles an
Error passed as the log *message* or as the top-level info object, never one
nested in metadata, which is the shape nearly every call site under `server/src`
uses.

`redact` now serialises an `Error` as `{ name, message, stack, cause, ...own
enumerable props }` before recursing, at the top-level `err`/`error` key and at
any nesting depth, inside arrays included. Own properties still go through the
existing key-based redaction, so `err.token` is `[REDACTED]` exactly as
`meta.token` would be, while `err.status` survives. The error view is recursed at
the *same* depth, so a logged error never costs a level of the depth-6 budget.

## Tests, and the mutations that prove they can fail

Every mutation below was applied to the implementation, the named suite was run,
and the file was restored.

| # | Mutation | Result |
|---|---|---|
| M1 | Delete the `input instanceof Error` branch in `redact` | `logger.test.ts` 6 failed / 21 passed |
| M2 | Drop `stack` from `errorShape` | 4 failed / 23 passed |
| M3 | `return errorShape(input)` without re-redacting own props | 1 failed / 26 passed (the `err.token` / `err.apiKey` arm) |
| M4 | Restore `sectionFailedWarning(group.label, String(err))` | 4 failed across the synthesis, guard and #225 suites |
| M5 | Make `publicDocWarnings` an unconditional pass-through | 5 failed across the unit and route suites |
| M6 | Ignore `detailSafe` (sanitise everything) | 1 failed — the "leaves a post-#67 warning untouched" arm |
| M7 | Drop the `detailSafe: true` marker from `sectionFailedWarning` | 2 failed |
| M8 | Point the guard's call-site scan at a name that does not exist | 1 failed — the "finds the call sites at all" arm |

M8 is the identity mutation: without it the guard would pass over an empty scan.

## Coverage

Measured with `vitest run --coverage` restricted to the three touched library
modules, over the same four suites before and after:

| | Statements | Branches | Functions | Lines |
|---|---|---|---|---|
| Before (`origin/main`) | 91.82% | 85.61% | 95.00% | 91.72% |
| After | 92.77% | 86.41% | 95.34% | 92.61% |

Both sides clear the project's 80% floor. `server/src/routes/generated-docs.ts`
and `holistic-synthesizer.ts` are also touched; their new lines are covered by
the route and synthesis suites and were proven failable by M4 and M5.

## Gate

`pnpm lint` (repo-wide, clean), `pnpm typecheck` (all 8 projects), `pnpm test`
(24,291 tests across 7 packages, 0 failed) all run green locally after
`pnpm --filter @metis/shared build`.

## Defect found along the way

`server/tests/unit/generated-docs-route.test.ts` stubbed
`grounding/degraded-warnings.js` as a **whole-module** mock exporting only
`deriveDocStatus`. Any other export the route reached for threw at request time
and surfaced as a 500 — the "wrong harness" shape: the fake diverged from the
module it stood in for, and nothing failed until a route actually needed a second
export. It is now a partial mock over `importOriginal`, keeping `deriveDocStatus`
as the spy the existing assertions use.

## Not done

- One pre-existing assertion was *changed*, not added:
  `holistic-synthesizer.grounding.test.ts`'s #225 case required the thrown text
  to appear in the warning message. That is exactly what #67 forbids, so it now
  asserts the section is named, the failure is surfaced, and the exception text
  is **absent**.
- No UI change. `DegradedWarningsBanner` renders `docDetail.data.warnings`
  straight from this endpoint, so it is covered by the server fix.
- No e2e work and nothing under `e2e/` or the `e2e` CI job was touched.
- `pnpm audit` reports 24 pre-existing advisories (4 low, 20 moderate, zero
  high/critical). This PR adds no dependencies, so none of them is new.


server/src/lib/docs-gen/generation-failure-message.ts
```diff
@@ -123,3 +124,40 @@ export function publicGenerationErrorMessage(
...
+ * post-#67 warning is passed through with its METIS-authored detail intact
+ * (the DB-schema synthesizer's "N of M tables described" text, for one). The
+ * flag is the discriminator precisely because no pre-#67 row can have it.
+ *
+ * Every other warning kind is returned untouched: none of them is ever built
+ * from an exception. The value is typed `unknown` because it comes off a Prisma
+ * `Json?` column, and anything that is not an array of objects — `null`, the
+ * pre-#252 string form — passes straight through.
+ */
+export function publicDocWarnings(warnings: unknown): unknown {
```
mgcronin, line 150: **Medium severity, high confidence on the code path — non-blocking.** The `warnings` column is closed, but the *pre-#252* storage location for the same warnings is not.

`publicGenerationErrorMessage` (line 118 of this file, unchanged) returns a non-`failed` row's `errorMessage` **verbatim**, and pre-#252 degraded rows stored `DocWarning[]` JSON in exactly that field. The UI still parses it: `resolveDocWarnings` in `ui/src/app/(authed)/projects/[id]/documentation/page.tsx:1121-1136` falls back to `JSON.parse(errorMessage)` whenever the structured column is empty, and renders whatever `message` strings it finds.

Those legacy warnings were built by the **old** `sectionFailedWarning(label, String(err))`, so a pre-#252 degraded row puts the same provider payloads, absolute paths and SQL text into the banner that this PR closes off in the `warnings` column. The doc comment on `publicGenerationErrorMessage` justifies the pass-through as "pre-#252 warning JSON the UI parses, which METIS wrote itself" — METIS *formatted* it, but it embedded the exception, which is the premise #67 disproves.

Suggestion (small): in `publicGenerationErrorMessage`, when `status !== "failed"` and the value parses as an array, route it through `publicDocWarnings` and re-stringify; otherwise leave it. Or file it as a follow-up if no pre-#252 rows are believed to exist in any deployed database — but that belief should be written down rather than implied.

server/src/lib/logger.ts
```diff
@@ -157,13 +157,41 @@ export function tokenCountMetaKeys(): ReadonlySet<string> {
...
+ *
+ * Own enumerable properties are spread on top (an `AIProviderError`'s `status`,
+ * a driver's `code`), and the caller walks the result through the same
+ * key-based redaction as any other object, so `err.token` redacts exactly as
+ * `meta.token` would.
+ */
+function errorShape(err: Error): Record<string, unknown> {
+  const out: Record<string, unknown> = { name: err.name, message: err.message };
+  if (err.stack != null) out.stack = err.stack;
+  if (err.cause != null) out.cause = err.cause;
```
mgcronin, line 178: **Low severity — non-blocking.** `errorShape` recovers `name`/`message`/`stack`/`cause`, but `AggregateError.errors` is own and **non-enumerable** (verified: `Object.getOwnPropertyDescriptor(new AggregateError([e],"m"), "errors").enumerable === false`), so it is dropped by both the explicit list and the `...err` spread — the same `{}` shape #68 exists to fix, one class down.

Any `Promise.any` rejection or an SDK that aggregates retry failures logs as an `AggregateError` with no sub-errors. One line covers it:

```ts
if (err instanceof AggregateError) out.errors = err.errors;
```

The sub-errors then recurse through `redact` like any other array, so their own key-based redaction still applies.

server/src/lib/logger.ts
```diff
@@ -157,13 +157,41 @@ export function tokenCountMetaKeys(): ReadonlySet<string> {
...
+}
+
 /**
  * Recursively walk a meta object and return a redacted clone.
  * Leaves the original untouched (winston shares meta across transports).
  */
 export function redact(input: unknown, depth = 0): unknown {
   if (depth > 6 || input == null) return input;
   if (Array.isArray(input)) return input.map((v) => redact(v, depth + 1));
+  if (input instanceof Error) {
```
mgcronin, line 189: **Praise + an informational note.** Recursing at the *same* depth is the right call and the comment says why — an error view is a view of one value, not a level of nesting. I re-applied the M1 mutation (deleting this branch) and confirmed `logger.test.ts` goes **6 failed / 21 passed**, so the branch is genuinely load-bearing.

Informational, out of this PR's scope: `docs/decisions/0008-redaction-sinks.md` treats three sinks as one family, and the other two still rebuild from `Object.entries` exactly as this one did — `server/src/lib/audit/audit-service.ts:194` and `server/src/lib/sandbox/audit/redact.ts:179`. An `Error` in an audit payload is still `{}` there. Those sinks **persist**, so "serialise the stack" is a decision rather than an obvious fix — which is an argument for a follow-up issue, not for widening this PR.

server/src/lib/docs-gen/generation-failure-message.ts
```diff
@@ -123,3 +124,40 @@ export function publicGenerationErrorMessage(
...
+  if (!Array.isArray(warnings)) return warnings;
+  return warnings.map((warning) => {
+    if (warning == null || typeof warning !== "object" || Array.isArray(warning)) return warning;
+    const w = warning as Record<string, unknown>;
+    if (w.kind !== "section-failed" || w.detailSafe === true) return warning;
+    const section = typeof w.section === "string" ? w.section : "Document";
+    const raw = typeof w.message === "string" ? w.message : "";
+    // Only the vocabulary's own output reaches the message; the legacy text is
+    // read solely to CLASSIFY it, never echoed.
+    const { message, detailSafe } = sectionFailedWarning(section, generationFailureMessage(raw));
```
mgcronin, line 160: nit (low confidence that it ever bites): the legacy message is classified *whole*, including the `Section "X" could not be generated: ` prefix, so a section label that itself contains a status-shaped token (`Section "HTTP 402 retries"`) would classify as a balance problem. Harmless — the output is still fixed-vocabulary text, only the choice of fixed string is wrong — but stripping the known prefix before classifying would remove the ambiguity, and the prefix is already reconstructed on the next line anyway.

server/tests/docs-gen-warning-detail.enumeration.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  }
+  return src.slice(openIdx);
+}
+
+/**
+ * Expressions that carry an exception's own text. `generationFailureMessage(err)`
+ * is deliberately NOT one of them: it consumes the error and returns a fixed
+ * string, which is exactly the required shape.
+ */
+const RAW_ERROR_EXPR = [
```
mgcronin, line 67: **Low severity — non-blocking.** Good guard: it re-derives from disk, it has an identity arm (`finds the call sites at all`) and a matcher arm, so neither an empty scan nor a deleted pattern leaves it green. That is better than most gates in this repo have shipped with.

Two boundaries worth stating in the doc comment, so the next reader does not over-trust it:
1. It is **textual and local to the argument expression** — `const detail = String(err); sectionFailedWarning(label, detail)` passes clean.
2. It guards **only `sectionFailedWarning`**. A future warning kind built from an exception (say a `section-truncated` carrying `err.message`) is outside the scan, even though `publicDocWarnings` only sanitises `kind === "section-failed"` and would pass it straight through.

Neither is a defect today — I audited the other builders: `db-schema-synthesizer.ts:289` interpolates `f.detail` only for `truncated-response` failures, whose detail comes from `describeTruncation` (METIS-authored), and `:303` joins `ProseFailureReason` slugs, never the `provider-error` detail that holds `err.message`. So the audit claimed in the PR body checks out.

server/src/routes/generated-docs.ts
```diff
@@ -406,6 +407,10 @@ export function generatedDocsRouter(): Router {
         ...doc,
         // #52 — never the raw exception text a pre-#52 row may still hold.
         errorMessage: publicGenerationErrorMessage(doc.status, doc.errorMessage),
+        // #67 — same rule for the DEGRADED path: a `section-failed` warning
+        // persisted before #67 appended up to 300 characters of `String(err)`
+        // to its message, and this column was returned verbatim.
+        warnings: publicDocWarnings(doc.warnings),
```
mgcronin, line 413: Confirmed by tracing rather than by trust: these two detail handlers are the **only** exits that carry the `warnings` column to a client. `GET /projects/:projectId/docs` (line 271) selects neither `warnings` nor `errorMessage`; the export handler (line 450) reads only `content`/`title`; the regeneration reads at 661/747/1082 never reach a response; `summarizeWarnings` is counts-only. The live socket path emits the object the now-fixed write path builds.

One thing nothing here proves, and it fails in the safe direction so it is only worth knowing: the route tests mock Prisma, so no test shows `detailSafe: true` surviving a real round-trip through the `Json?` column. If it were ever dropped in storage, a post-#67 warning would simply be re-sanitised on read — a lost METIS-authored detail, never a leak.

mgcronin (commented):
## Code Review — PR #80 (#67, #68)

Recommendation: merge with follow-ups — 0 blocking findings, 5 non-blocking.

This is **automated analysis published as a comment, not a human sign-off**: the
authenticated `gh` identity is also the PR author, so GitHub cannot record an
independent approval or change request from it. A human still owns the merge decision.

### Requirements: PASS
**#67** — a thrown section error now maps through `generationFailureMessage` before the
warning is built (`holistic-synthesizer.ts:3114`), and persisted pre-#67 warnings are
re-derived on read in both `GET /:docId` and `PATCH /:docId`. Both acceptance criteria are
covered by tests that I ran and mutated: the secret-detail case and the 402-stays-
recognisable case.
**#68** — `redact()` serialises an `Error` at the top-level `err`/`error` key, nested, and
inside arrays; `err.token` still redacts and `err.status` survives. Both acceptance
criteria covered, including one arm through the wired winston instance (a real read-back,
not an assertion against the object the test just built).

### Verification I ran myself (not taken from the PR body)
- Six touched suites, unmutated: **6 files / 137 tests passed**.
- **M4 re-applied** (`sectionFailedWarning(group.label, String(err))`): **4 failed** across
  the redaction, #225 grounding and enumeration-guard suites. Restored.
- **M1 re-applied** (delete the `instanceof Error` branch in `redact`): **6 failed / 21
  passed** in `logger.test.ts` — exactly the tally claimed. Restored.
- **M5 re-applied** (`publicDocWarnings` → pass-through): **5 failed** across the unit and
  route suites. Restored. `git status` clean after each.
- Coverage over the touched library modules with their own suites included:
  **95.6% statements / 88.9% branches / 97.7% functions / 96.0% lines**, against the
  project's 80% floor in `CLAUDE.md`.
- `eslint` on the 11 changed files with `--max-warnings=0`: clean.
  `tsc --noEmit -p server/tsconfig.json`: clean.
- CI on the exact head SHA `49d20f4`: 16 checks, 14 pass, `e2e` and `windows` **skipping**,
  zero pending, zero fail on a plain re-read. `api` pass (28m4s) with `api-outcome` pass,
  `postgres-adapter` pass (7m32s) — no supersede, no teardown flake.

### Security: 1 residual finding, no new exposure
The write path and both detail-endpoint read paths are closed. I traced every other exit
of the `warnings` column myself: `GET /projects/:id/docs` (list) selects no `warnings` and
no `errorMessage`; the export handler reads only `content`/`title`; `summarizeWarnings` is
counts-only; the live `job:doc-section` socket emits the same now-sanitised object. The one
gap is the **legacy `errorMessage` warnings JSON** (comment inline) — the same exposure
class, in the pre-#252 storage location the UI still parses.
No injection, no new dependency, no secret, no scope/tenancy change (`projectId` filters
are untouched on every query the PR goes near).

### Design: good
The `detailSafe` discriminator is a sound choice — no pre-#67 row can carry it, so the read
path is exact rather than heuristic — and the author correctly identified that the marker
alone would be fail-open for a *new* call site and wrote a call-site guard **with an
identity arm and a matcher arm**. That is the failure mode most guards in this repo have
shipped with (#1168, #1192), and it is handled here.
`generation-failure-message.ts` importing `degraded-warnings.ts` introduces no cycle
(`degraded-warnings.ts` has zero imports).

### Tests: sufficient, and falsifiable
The redaction test drives the real synthesis loop rather than calling the warning builder,
and the logger test reads back through the wired winston transport. The whole-module
`vi.mock` → partial-mock repair in `generated-docs-route.test.ts` is a real defect fixed,
not diff noise, and it is persisted to `.claude/agent-memory/code-issue/` and committed
(1,074 B index, under the 17,500 B budget).
One pre-existing assertion was **changed** rather than added
(`holistic-synthesizer.grounding.test.ts:718`) — I checked it: the old assertion required
the thrown text to be present, which is precisely what #67 forbids, and the replacement is
stronger (names the section, asserts the exception text is absent).

### Documentation: up to date
Two changelog fragments under `.changes/unreleased/`, per `CLAUDE.md` — `CHANGELOG.md` not
touched. No version bump. No architecture change to document. Nothing under `e2e/` or
`.github/workflows/`: the scope fence is honoured.

### Scope of this review
Read all 15 changed files in full plus the surrounding read/write paths in
`generated-docs.ts`, `db-schema-synthesizer.ts`, the two other redaction sinks and the UI
banner. Ran the suites, mutations, lint, typecheck and coverage listed above. Did not
exercise the running application, and did not run the full 24k-test fan-out locally — CI
did, green, on this head SHA.

server/src/lib/docs-gen/grounding/degraded-warnings.ts
```diff
@@ -83,6 +83,19 @@ export interface DocWarning {
...
+   * wrote), rather than an exception's own message.
+   *
+   * It exists so the READ path can tell a post-#67 warning from one persisted
+   * before it. A `section-failed` warning without this flag may still carry up
+   * to 300 characters of raw `String(err)` — provider response bodies, absolute
+   * paths, SQL text — so `publicDocWarnings` re-derives its detail through the
+   * fixed vocabulary before any of it reaches a client. Absent on every other
+   * warning kind, none of which is ever built from an exception.
+   */
+  detailSafe?: boolean;
```
mgcronin, line 98: Nice: `detailSafe` is a genuinely exact discriminator rather than a heuristic — no row written before this change can carry it — and the doc comment records *why* it exists, which is the part that usually rots. Pairing it with the call-site enumeration guard closes the fail-open path a marker alone would have opened.

Issue #81, opened 2026-09-22
Test-coverage budget endpoint reports the default cap, not the one the run used


mgcronin:
The per-run cap a coverage run actually ran under is never written anywhere, so the
budget endpoint reports a different number.

`runCoverageScoring` takes the cap from `deps.budgetCents`
(`server/src/lib/testcoverage/coverage-service.ts`), which the task-runner passes through
from `TestCoverageRunnerDeps.budgetCents` (`task-runner.ts:57,142`). The tracker holds it
in memory for the life of the run, and `flush()` persists only
`tokenCostCents / embeddingTokens / judgeTokens / suggestionTokens` — `TestCoverageRun`
has no column for the limit (`server/prisma/schema.prisma`, `model TestCoverageRun`).

The read path then invents one:

```ts
// server/src/routes/test-coverage.ts:322
const view = await readBudget(runId, { budgetCents: DEFAULT_BUDGET_CENTS });
```

`DEFAULT_BUDGET_CENTS` is read from `TESTCOVERAGE_BUDGET_CENTS` **at module import**
(`cost-tracker.ts`). So for any run whose cap was not that value — a caller passing
`budgetCents`, or an operator who changed the env var and restarted — the budget tile's
`limitCents` and `remainingCents` are wrong, and `remainingCents` can read healthy on a run
that stopped at `budget_exceeded`. `usedCents` and the token breakdown are correct; only the
cap and everything derived from it are not.

This is the write/read shape: the write reports success, and the read cannot see what was
written because it was never written at all. Found while working #72/#73/#77 (PR that fixes
them). Out of scope there: the fix needs a migration.

## Acceptance criteria
- [ ] The cap a run executed under is persisted on `TestCoverageRun` (migration + backfill
      default), written at the same point the run's other budget columns are.
- [ ] `readBudget` returns that stored cap, falling back to `DEFAULT_BUDGET_CENTS` only for
      rows written before the migration.
- [ ] A test starts a run with a non-default `budgetCents`, reads the budget back through
      `readBudget`, and sees the same `limitCents` — red before the fix.
- [ ] A route-level test covers `GET /api/projects/:id/test-coverage/runs/:runId/budget`.
      It has no test today; `readBudget` is only exercised through a unit-test double, so
      nothing checks the endpoint's own contract.

## Also here, while the file is open
`tokenCostCents` is documented in the schema as "Sum of token costs (cents * 100)". It holds
[...]

PR #82, opened 2026-09-22
fix: Resolve #72, #73, #77 — in-loop embedding usage on the budget; the judge's cost guard covered; an unpriced embedder no longer ends a run

### LINKED ISSUE #72: Test-coverage in-loop embedding calls (judge and suggestion cache keys, dedup) are never recorded

### LINKED ISSUE #73: No test covers coverage-service passing its cost guard to the judge

### LINKED ISSUE #77: Test-coverage cost follow-ups from #74: untested embedding arm, unlisted embedder blocks budgeted runs
Non-blocking findings from the review of #74 (https://github.com/openzigs/metis/pull/74#pullrequestreview-5281324128), which fixed #57 and #58.

- [ ] **The test-case embedding arm is untested by construction.** Deleting the `cost.record({ phase: "embedding", … caseEmb … })` call in `server/src/lib/testcoverage/coverage-service.ts` leaves all 432 tests green, because the #58 service fixture has no test cases — so its `toHaveLength(1)` assertion passes either way. On a cloud embedder that silent loss under-charges the budget and stops `exceeded()` firing. Add a fixture with test cases and assert both embedding records.
- [ ] **Embedding rate rows are exact model ids, while the provider keys accept anything.** `bedrock`, `bedrock-sdk` and `openai` also serve Titan V1, Cohere, Azure deployment names and self-hosted OpenAI-compatible endpoints. Since embedding is a run's first recorded usage, an unlisted deployment makes `exceeded()` true before the judge starts, so every such run stops immediately. Decide whether an unpriced embedder should block a budgeted run, or be treated differently from unpriced generation.

The remaining two findings are in the review.

mgcronin:
Closes #72
Closes #73
Closes #77

Three follow-ups to PR #74, all on the test-coverage run budget. They are one PR
because they collide: #72 makes the judge and the suggestion phase record their
own embedder calls, and under #77's old rule an unpriced embedder would then have
stopped the judge after its first batch rather than merely before it started.

## #72 — the in-loop embedder calls reach the budget

**Before:** only the match phase's two `embed()` calls were recorded. The judge
embedded every batch prompt for its semantic-cache key, and `generateSuggestions`
embedded every cluster prompt plus every suggestion's text for dedup. None of
that reached `CoverageCostTracker`, the run budget, or `ai_token_usages`. With a
local embedder it cost nothing; with `bedrock`, `bedrock-sdk` or `openai` it was
real spend nothing could see.

**Now:** every embedder call is recorded through the phase's existing cost guard,
under `getEmbedder().key` read *after* `embed()` (a failed backend may have been
swapped for the hash stub) and the model `embed()` itself reported — the same
rule #58 set for the match phase. `JudgeBudgetGuard` and `SuggestionBudgetGuard`
take the embedding arm of `CoverageUsage` as well as their own phase.

The cache-key embedding is recorded on a cache **hit** too: a warm cache skips the
model call, not the embedding — the key is what the lookup needs.

`estimateTokens` moved out of `coverage-service.ts` to
`cost-tracker.ts:estimateEmbeddingTokens`, so all five call sites bill on one
basis instead of the service having a private copy.

## #73 — the service's cost guard for the judge is covered

**Before:** deleting `cost` from the `judgeAmbiguous(...)` call left all 139 tests
green. Without it the judge takes the legacy path: one total recorded at the end
through the global `TokenTracker`, and no budget check between batches. A run
could make every judge call past its cap with the service tests still passing.

**Now:** a service test runs a fixture of three requirements against four test
cases whose vectors are pinned into the matcher's AMBIGUOUS band (cosine 0.70,
where the hybrid BM25 rule cannot apply) — twelve AMBIGUOUS cells, two judge
batches of eight. One batch of 100 prompt + 50 completion Bedrock Haiku tokens
rounds up to exactly the one-cent cap, so the second batch must not run. The test
pins `batches === 1`, `modelCalls === 1`, `budgetExceeded === true` and the
`judge` done event.

The fixture needed a way to choose vectors, so the file's embedder mock now takes
prefix-pinned vectors and a settable `key`/`model`. That same seam is what makes
the #77 tests below possible.

## #77 — an unpriced embedder no longer ends a budgeted run

**Finding 1 — the test-case embedding arm was untested by construction.** The #58
service fixture has no test cases, so its `toHaveLength(1)` on the embedding rows
held whether or not the case arm recorded anything. The new fixture has four test
cases and asserts both match-phase arms are billed, by the token count of the
exact texts the embedder was handed. Deleting the case arm is mutation M9 below —
one red test now, zero before.

**Finding 2 — the decision.** `bedrock`, `bedrock-sdk` and `openai` are provider
keys that accept any model: Titan V1, Cohere, an Azure deployment name, a
self-hosted OpenAI-compatible endpoint. The rate rows are exact model ids, so all
of those are unpriced — and because embedding is a run's **first** recorded usage,
`exceeded()` was true before the judge started and every such run stopped
immediately.

**Decided: unpriced embedding is reported, not blocking. Unpriced judge and
suggestion usage still fails closed (#43).** They are not alike:

- Embedding spend is bounded, input-only, and already incurred by the time it is
  recorded. At the dearest published rate on the table (ada-002, $0.10/MTok) a
  100-req / 200-test run's ~12.5k tokens is under a fifth of one cent against a
  20-cent cap. Refusing the run buys none of it back.
- LLM spend is the open-ended part the cap exists to control, and the part the run
  can still decline to make. That one still stops.

The uncertainty is surfaced rather than hidden: `unpricedTokens` still counts every
unpriced token, so the budget line still reads `$x + N unpriced tokens` and
`usedCents` reads as the lower bound it is. `CoverageBudgetView` gains
`unpricedEmbeddingTokens` and `unpricedLlmTokens`; `readBudget` computes the same
split from the persisted rows' `agentStep`, so the read path answers the same
question the in-memory one does. `MODEL_PRICES` with key `embed:<backend>:<model>`
still puts a model back on the cap. USER_GUIDE says all of this.

No price row was added: I am not inventing prices for models I did not verify.

## Tests that fail before the fix

The mutation table is the evidence, not a green suite. Each was applied alone and
restored, over `tests/lib/testcoverage/` + `src/lib/finops/` + `tests/lib/finops/`
(442 tests). **All ten went red.**

| # | Mutation | Went red |
|---|---|---|
| M1 | Judge never records its cache-key embedding | 3 |
| M2 | Generator never records the cluster-prompt embedding | 4 |
| M3 | Generator never records the dedup embedding | 3 |
| M4 | Judge records the CONFIGURED model, not the one `embed()` reported | 1 |
| M5 | Service does not pass `cost` to `judgeAmbiguous` — **the #73 survivor** | 3 |
| M6 | `exceeded()` fails closed on unpriced embedding again | 4 |
| M7 | `canAfford()` refuses on unpriced embedding again | 1 |
| M8 | `readBudget` does not filter the embedding aggregate by phase | 1 |
| M9 | Service drops the test-case embedding record — **the #77 untested arm** | 1 |
| M10 | `view()` hides the unpriced embedding share | 4 |

M5 and M9 are the two survivors #73 and #77 were filed for. Both are killed.

## Coverage (touched source files, lines / branches)

Measured over the same three suites, at `origin/main` and at this branch.

| File | Before | After |
|---|---|---|
| `testcoverage/judge.ts` | 97.46 / 88.00 | **100 / 96.00** |
| `testcoverage/coverage-service.ts` | 87.90 / 66.66 | **97.52 / 80.00** |
| `testcoverage/suggestion-generator.ts` | 95.52 / 81.08 | **97.08 / 83.78** |
| `testcoverage/cost-tracker.ts` | 100 / 92.30 | **100 / 92.85** |
| `finops/provider-rates.ts` (untouched) | 97.40 / 96.59 | 97.40 / 96.59 |

The floor is 80% lines; every touched file clears it, and `coverage-service.ts`
gains nearly ten points — the #73 test is the first to drive its judge branch.

## Gate (run locally, in the worktree)

- `pnpm lint` — **exit 0**. 4,055 tracked files scanned; 0 of 43 private terms found.
- `pnpm typecheck` — **exit 0**, all 8 workspace projects.
- `pnpm test` — **exit 0**. server 17,609; ui 3,373; scripts 2,367; shared 698;
  embeddings-svc 147; copilot-svc 78; ui-kit 7. No package failed and none was
  skipped over.
- `node scripts/verify-changelog-fragment.mjs --base origin/main` — passes, with the
  three fragments below.
- `pnpm agents:verify` — exit 0 (the memory index is 1,079 B of its 17,500 budget).

## Security

- No dependency added or bumped, so no audit was run.
- OWASP pass over the diff: the two new Prisma filters (`agentStep` equality and an
  `in` over two constants) take no user input; no query is string-built; no shell;
  no deserialization; no secret. The values newly persisted as `provider`/`model`
  come from the embedder backend, not from a request. `MODEL_PRICES` stays
  admin-set and schema-validated.

## Write/read check

- The embedding rows are asserted through the mocked `aITokenUsage.create`, by the
  token count of the exact texts the embedder was handed — so a dropped arm cannot
  hide behind another arm's row.
- `readBudget`'s split is asserted against the `where` clauses it actually issues,
  and M8 proves that assertion can fail.
- **Found, filed as #81, not fixed here:** the cap a run executed under
  (`deps.budgetCents`) is never persisted — `TestCoverageRun` has no column for it —
  so the budget endpoint always reports the process-wide `DEFAULT_BUDGET_CENTS`.
  For any run started with a different cap, `limitCents` and `remainingCents` are
  wrong and `remainingCents` can read healthy on a run that stopped at
  `budget_exceeded`. It needs a migration, so it is out of scope here. #81 also
  records that `GET …/runs/:runId/budget` has **no route-level test** — `readBudget`
  is exercised only through a unit-test double — and that the schema comment on
  `tokenCostCents` ("cents * 100") contradicts the code, which writes cents.
- A record with no provider or model is counted in memory but never persisted, so
  `readBudget` cannot see it. Pre-existing and unreachable today (`Embedder.key` is
  always a string); noted rather than worked around.

## Deliberately deferred

- `unpricedEmbeddingTokens` / `unpricedLlmTokens` ship **unconsumed by the UI**. The
  test-coverage budget tile reads `unpricedTokens`, which still totals both and is
  still correct. Splitting the tile is UI work none of the three issues asked for.
- No new embedding price rows. Titan V1, Cohere on Bedrock and Azure deployment
  names stay unpriced by design; #77 asked for the policy, not for prices I would
  have had to guess.
- The remaining two findings from the #74 review, and the #59 review's synthetic
  `AISession` / `recordAndFlush` items, are untouched.

---

## Review round 2 — the blocking finding, fixed at the source

The review's blocking finding was right and its cheapest remedy was declined. It offered
"narrow the changelog and USER_GUIDE sentences to the judge/suggestion loops and file the
index-phase follow-up". That would have closed #72 — whose AC1 says *every* embedder call
a **run** makes — with the run's largest embedding consumer still off-budget. The wiring
turned out to be contained, so the claim was made true instead.

**What was wrong:** `task-runner.ts` drives `import` → `index` → `match`/`judge`/`suggest`
→ `score`. `TestCoverageIndexer.index()` embeds every test-case text (`indexer.ts:102`)
and every step text (`indexer.ts:111`). `CoverageCostTracker` was constructed *inside*
`runCoverageScoring`, which the `index` phase runs before. On a cold 200-test corpus those
two batches are larger than every cache-key and dedup embedding this PR added.

**What changed:** the task-runner builds ONE tracker per run and threads it into both
`indexer.index(projectId, indexable, { cost })` and `runCoverageScoring(…, { cost })`.
`CoverageServiceDeps.cost` is optional, so calling the service on its own still builds its
own. The runner also flushes: with no `caller` wired the service never runs at all, and the
index phase's usage would otherwise never be persisted. The flush is `.catch`-logged —
accounting must not be the reason a completed run reads as failed.

Every `embed()` call in `server/src/lib/testcoverage/` is now billed. The four sites are
`indexer.ts` (×2), `coverage-service.ts` (×2), `judge.ts` (×1), `suggestion-generator.ts` (×2).

### The other findings

| # | Finding | Action |
|---|---|---|
| 2 | `ARCHITECTURE.md` still said the budget "treats unpriced usage as exceeded" | **Taken.** Both passages corrected, plus a new paragraph naming the phase list and where the tracker is built. |
| 3 | #72 AC2 proven only compositionally | **Taken.** Two direct tests, below. |
| 4 | `readBudget`'s LLM aggregate was an allowlist where `view()` uses an `else` | **Taken.** It now aggregates the session's unpriced total and subtracts the embedding share. |
| 5 | Two view fields with no consumer and no follow-up | **Taken.** #92 filed; both declarations cite it. |
| 6 | One extra `ai_token_usages` insert per judge batch (informational) | **Not changed.** It is the point: a row per embedder call is what makes the spend auditable. The write is fire-and-forget behind `ensureSession`, so it adds no latency to the batch. |
| 7 | `vi.unstubAllEnvs()` in two test bodies leaks on failure | **Taken.** Moved to `afterEach` in `judge.test.ts`, `suggestion-generator.test.ts` and the new stub site in `coverage-service.test.ts`. |

**Finding 3 in detail.** The published OpenAI rate is fractions of a cent on a fixture this
size, so cents cannot show the in-loop share moving the budget. Two tests, therefore:

- *published rate* — `embed:openai` / `text-embedding-3-large`: all six embedding rows priced
  at exactly `tokens × $0.13/MTok`, `unpricedEmbeddingTokens === 0`, the four in-loop rows are
  the majority of the run's embedding tokens, and `usedCents === ceil(Σ estimatedCostUsd × 100)`.
- *administrator's price* — the same run with a `MODEL_PRICES` key
  `embed:openai:text-embedding-3-large` at a price that registers: the four in-loop calls are
  worth **hundreds of cents** of the run's reported spend, which is the `embed:<backend>:<model>`
  path #77 documents and nothing exercised end-to-end.

Both are identified by set-difference on the token counts of the exact texts the embedder was
handed, so they do not depend on row ordering.

## Round-2 mutations (each applied alone, restored, and the restore asserted)

| # | Mutation | Went red |
|---|---|---|
| M11 | `indexer` drops the CASE-batch embedding record | 2 |
| M12 | `indexer` drops the STEP-batch embedding record | 1 |
| M13 | `task-runner` does not hand its tracker to the index phase | 3 |
| M14 | `task-runner` never flushes the run tracker | 1 |
| M15 | `coverage-service` ignores the injected tracker and builds its own | 1 |
| M16 | `readBudget` back to the two-value LLM allowlist | 2 |

Re-verified from round 1, against the changed service: **M1′** (judge never records its
cache-key embedding) → 2 red, *both of them the new AC2 tests* — they are not vacuous.
**M5′** (the #73 survivor) → 4 red. **M9′** (the #77 survivor) → 3 red.

## Coverage after round 2 (lines / branches)

| File | After |
|---|---|
| `testcoverage/cost-tracker.ts` | 100 / 92.85 |
| `testcoverage/judge.ts` | 100 / 96.00 |
| `testcoverage/indexer.ts` | 98.07 / 100 |
| `testcoverage/coverage-service.ts` | 97.52 / 80.51 |
| `testcoverage/suggestion-generator.ts` | 97.08 / 83.78 |
| `testcoverage/task-runner.ts` | 95.12 / 85.71 |

Every touched file clears the 80% line floor. `indexer.ts` and `task-runner.ts` are newly
touched in this round; their pre-change baselines were not re-measured.

## Round-2 gate

- `pnpm lint` — exit 0 (4,059 files; 0 of 43 private terms).
- `pnpm typecheck` — exit 0, all 8 projects.
- `pnpm test` — server 17,618 passed / 11 skipped; ui 3,373 passed / 1 skipped;
  scripts 2,367; shared, embeddings-svc, copilot-svc, ui-kit all green. One run out of three
  showed `scripts` at `Test Files 1 failed` with no file named in the fan-out output; the
  package passed in isolation and on both other full runs, and nothing in `scripts/` is
  touched by this change.
- `npx vitest run tests/lib/testcoverage tests/lib/finops src/lib/finops` — 40 files / 451 tests.
- `node scripts/verify-changelog-fragment.mjs --base origin/main`, `pnpm agents:verify`,
  `pnpm skills:verify` — all pass.

## Round-2 security

No dependency added or bumped. The one new query shape is `readBudget`'s session-wide unpriced
aggregate: `where: { sessionId, estimatedCostUsd: null }`, where `sessionId` is
`testCoverageRun:<runId>` built by `coverageSessionId` — a namespace only this tracker writes,
Prisma-parameterised, no user-supplied string interpolated. The index phase records
`provider`/`model` read from the embedder backend, not from a request. No shell, no
deserialization, no new auth or scoping surface.

## Write/read check, round 2

`flush()` is still the only writer of the columns `readBudget` reads, and it runs once at the
end of a run — so during the run the budget endpoint reports zeros. That is pre-existing for
the match phase, but this PR makes more real spend invisible mid-run, since the index phase is
now billed and on a cold corpus is the largest part of it. Wrong *time*, not wrong storage:
`view()` has the live numbers the whole time and nothing exposes them. Filed as a comment on
**#81**, which already owns this endpoint's other two defects and asks for the route-level test
a fix would want.


.changes/unreleased/72-in-loop-embedding-usage.md
```diff
@@ -0,0 +1,11 @@
+---
+issue: 72
+section: Fixed
+---
+
+- Every embedder call a test-coverage run makes now reaches the run budget, not just the two in the
```
mgcronin, line 6: **BLOCKING (requirements / documentation accuracy).** "Every embedder call a test-coverage run makes now reaches the run budget" is not true yet, and neither is USER_GUIDE's "every call a run makes" (line 4423).

`task-runner.ts` runs an `index` phase inside every coverage run, before `runCoverageScoring`. It calls `TestCoverageIndexer.index()`, which embeds **every test-case text** (`indexer.ts:102`) and **every step text** (`indexer.ts:111`). The cost tracker is constructed inside `runCoverageScoring`, so none of that reaches it. On the cold 200-test corpus the 20c budget AC is written against, those two batches are far more tokens than all the cache-key and dedup embeddings this PR adds — so on a cloud embedder the largest share of embedding spend is still invisible to the budget and still absent from `ai_token_usages`. (Warm re-runs skip it: `index()` returns early when every `contentHash` is already indexed, so this is a cold-run cost specifically.)

I am not asking you to wire the indexer up here — that means threading a tracker through `TestCoverageRunnerDeps` and is a fair second PR. What blocks is the combination: merging closes #72, whose AC1 says *every* embedder call, while the changelog and the user guide tell an operator the budget now sees all of it. That is how a real gap stops being looked at.

Cheapest way to clear this:
1. Narrow this fragment and USER_GUIDE:4423 to what shipped — the judge's per-batch cache keys and the suggestion phase's cluster prompts and dedup.
2. File the index-phase follow-up and reference it, the way #81 was filed from this same run.

(Worth noting in that follow-up: `index()` also issues one `store.search` per case in a loop, `indexer.ts:89` — pre-existing, not yours, but the same function.)

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -270,9 +309,27 @@ export class CoverageCostTracker {
...
+   * OpenAI-compatible endpoints, so failing closed here meant every such
+   * deployment got zero coverage runs — and, with #72 recording the in-loop
+   * calls too, would have stopped the judge after its first batch. The
+   * uncertainty is still surfaced: `unpricedTokens` counts those tokens, so
+   * `usedCents` reads as the lower bound it is, and an administrator can price
+   * the model with a `MODEL_PRICES` key `embed:<backend>:<model>`.
    */
   exceeded(): boolean {
-    return this.unpricedTokens > 0 || this.usedCents >= this.limitCents;
+    return this.unpricedLlmTokens > 0 || this.usedCents >= this.limitCents;
```
mgcronin, line 336: **Medium — `docs/ARCHITECTURE.md` now contradicts this line in two places.** USER_GUIDE was updated; the architecture doc was not.

- `ARCHITECTURE.md:5849` — "…their per-run budget treats unpriced usage as exceeded (#43)."
- `ARCHITECTURE.md:~6221` (Cost Telemetry #878) — "…so a run stops part-way once the cap is reached **or a call reveals an unpriced model**."

Both are now false for the embedding phase, and the second sits in the section that enumerates exactly this behaviour. The same section's embedding bullet is also the natural place to say that the judge's and generator's in-loop calls are billed (#72). Two short edits.

The decision itself reads well and the rationale in this docstring is the right level of detail — bounded, input-only, already incurred, versus the open-ended spend the cap exists to control. Keeping #43 fail-closed for the LLM arms and proving it with the "an unpriced JUDGE model still stops the run, even behind a priced embedder" control test is the right shape.

server/tests/lib/testcoverage/coverage-service.test.ts
```diff
@@ -440,11 +464,228 @@ describe("runCoverageScoring", () => {
...
+    const billed = embedding.map((r) => r.totalTokens);
+    expect(billed).toContain(embedTokens(reqTexts));
+    expect(billed).toContain(embedTokens(caseTexts));
+    // The run row's total is every embedder call, not just the match phase.
+    expect(state.run.embeddingTokens).toBe(
+      embedState.calls.reduce((n, texts) => n + embedTokens(texts), 0),
+    );
+  });
+
+  it("an unpriced cloud embedder no longer stops a budgeted run (#77)", async () => {
```
mgcronin, line 732: **Low (tests).** #72's second acceptance criterion is "A test with a **priced cloud embedder** shows these calls on the run budget." What exists proves it in two halves: this file proves the in-loop calls reach the tracker with the right embedder key and model, and `cost-tracker.test.ts:341` proves `embed:bedrock:amazon.titan-embed-text-v2:0` and `embed:openai:text-embedding-3-large` price correctly. Nothing joins them.

This test already has the seam. A sibling that sets `embedState.key = "openai"; embedState.model = "text-embedding-3-large"` and asserts `report.cost.usedCents > 0` with `unpricedTokens === 0` would close the AC literally, and would catch a class the current tests cannot: an in-loop record that reached the tracker under the wrong provider namespace (e.g. `openai` rather than `embed:openai`, which is the mistake #58 existed to prevent). It would also be the only test where embedding spend alone can move `exceeded()`.

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -312,17 +369,34 @@ export async function readBudget(
...
+  const [unpricedEmbedding, unpricedLlm] = await Promise.all([
+    db.aITokenUsage.aggregate({
+      where: { sessionId, estimatedCostUsd: null, agentStep: "testcoverage.embedding" },
+      _sum: { totalTokens: true },
+    }),
+    db.aITokenUsage.aggregate({
+      where: {
+        sessionId,
+        estimatedCostUsd: null,
+        agentStep: { in: ["testcoverage.judge", "testcoverage.suggestion"] },
```
mgcronin: **Low (design).** The LLM aggregate is an allowlist of two `agentStep` values, but `unpricedTokens` is then derived as `embedding + llm`. Today that is exact — `cost-tracker.ts:224` and `judge.ts:364` are the only writers for a `testCoverageRun:` session, and they emit only these three steps, so nothing is lost. I checked.

The asymmetry is with the in-memory path, which routes *everything* that is not `embedding` into `unpricedLlmTokens` via an `else`. Add a fourth phase later and the two views disagree: `view()` counts it, `readBudget()` drops it from both the split *and* the total, so the persisted "+ N unpriced tokens" line silently under-reports.

Cheap way to make the total filter-independent: keep the embedding aggregate as-is, and compute the total from an unfiltered `estimatedCostUsd: null` aggregate, with `unpricedLlmTokens = total - embedding`. Then a new phase lands in the LLM arm by default — which is also the fail-closed direction.

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -53,11 +63,20 @@ export interface CoverageBudgetView {
...
-   * budget refuses further LLM work.
+   * Every token this run spent on a model METIS has no price for, across all
+   * phases. NOT in `usedCents`; non-zero means `usedCents` is a lower bound.
    */
   unpricedTokens: number;
+  /**
+   * #77 — the embedding share of {@link unpricedTokens}. Reported, but it does
+   * not stop the run: see {@link CoverageCostTracker.exceeded}.
+   */
+  unpricedEmbeddingTokens: number;
```
mgcronin, line 78: **Low.** `unpricedEmbeddingTokens` and `unpricedLlmTokens` are new on the public view with no production consumer: the route hands the view straight out, and the test-coverage budget tile (`ui/.../test-coverage/page.tsx:707`) reads `unpricedTokens`. I checked the tile's wording and agree the deferral is safe — it renders a neutral "+ N unpriced tokens", not a "run blocked" message, so an unpriced embedder does not now read as a stop.

Note the private fields *are* load-bearing (`exceeded()`, `canAfford()`); it is only the two view fields that are unproven. Unconsumed API surface is where defects hide, because the first consumer is what surfaces them. The PR body defers this deliberately, but nothing tracks it — a one-line follow-up issue for splitting the tile would keep it from becoming permanent.

server/src/lib/testcoverage/judge.ts
```diff
@@ -269,8 +277,18 @@ export async function judgeAmbiguous(
...
+    const embedded = await embedder.embed([userPrompt]);
+    const cacheKey = embedded.vectors[0];
+    // #72 — the cache-key embedding is a real embedder call, made whether or
+    // not the lookup then hits. Recorded under the embedder that ran (read
+    // AFTER embed: a failed backend may have been swapped for the hash stub)
+    // and the model IT reported, the same way the match phase is (#58).
+    options.cost?.record({
+      phase: "embedding",
+      embedder: embedder.key,
+      modelId: embedded.model,
```
mgcronin, line 289: Good call recording this on a cache **hit** as well — the key is what the lookup needs, so the embedder ran and billed either way, and the dedicated cache-hit test is what makes that falsifiable. Reading `embedder.key` and `embedded.model` *after* `embed()` is also right, and the "records the CONFIGURED model" mutation proves the test notices.

Informational, for sizing rather than change: each of these is a separate `ai_token_usages` insert via `recordAndFlush` (`cost-tracker.ts:236`). The judge phase's persisted row count doubles (one embedding + one judge row per batch) and the suggestion phase gains two per cluster; all of them are awaited in `flush()`'s `inflight` set. On a large run that is more round trips than before. Fine at current volumes — worth remembering if per-run row counts ever show up in the usage tables' query times.

server/tests/lib/testcoverage/judge.test.ts
```diff
@@ -363,4 +365,65 @@ describe("judgeAmbiguous", () => {
...
+    const cold = guard();
+    await judgeAmbiguous(pairs, { ...opts, cost: cold.cost });
+    const warm = guard();
+    const second = await judgeAmbiguous(pairs, { ...opts, cost: warm.cost });
+
+    expect(second.cacheHits).toBe(1);
+    expect(second.modelCalls).toBe(0);
+    expect(warm.recorded.filter((r) => r.phase === "judge")).toHaveLength(0);
+    expect(warm.recorded.filter((r) => r.phase === "embedding")).toHaveLength(1);
+    vi.unstubAllEnvs();
```
mgcronin: nit: `vi.unstubAllEnvs()` at the end of the test body only runs when the test passes. If an assertion above it throws, `SEMANTIC_CACHE_ENABLED=1` leaks to every later test in the file — and the `beforeEach` resets the cache singleton but not the env, so the leak is invisible until a semantics change makes a cached path behave differently. `afterEach(() => vi.unstubAllEnvs())` is leak-proof. Same shape in `suggestion-generator.test.ts:439`.

mgcronin (commented):
## Code Review — PR #82 (#72, #73, #77)

Recommendation: do not merge — 1 blocking finding (cheap to clear: two sentences and a follow-up issue, not code).

This is automated analysis published as a comment, not a human sign-off: the reviewing identity is also the PR author, so GitHub cannot record an approval or a change request from it. A human owns the merge decision.

### Requirements: PARTIAL
- **#72 AC1** — "Every embedder call made by a coverage run is recorded through `CoverageCostTracker`": **not met.** The judge and suggestion loops are covered, but the run's `index` phase (`TestCoverageIndexer.index`, `indexer.ts:102,111`, called from `task-runner.ts` inside every run) still embeds every test-case text and every step text unrecorded — on a cold run that is the bulk of a run's embedding spend. The changelog fragment and USER_GUIDE assert completeness anyway. See the blocking comment.
- **#72 AC2** — "a test with a priced cloud embedder shows these calls on the run budget": met only compositionally (in-loop record → unit test; `embed:<key>:<model>` pricing → tracker test). No single test drives the in-loop path with a priced embedder. Low.
- **#73 AC1/AC2** — met. Verified by re-running the mutation myself (below).
- **#77 finding 1 and finding 2** — met; the decision is implemented, tested both ways, and documented in USER_GUIDE.

### Verification I ran myself (not taken on trust)
| Check | Result |
|---|---|
| `pnpm lint` (worktree, PR head) | exit 0 |
| `pnpm typecheck` | exit 0, 8 projects |
| `npx vitest run tests/lib/testcoverage tests/lib/finops` | 32 files / 313 tests passed |
| `pnpm agents:verify` | exit 0 (index 1,079 B) |
| `node scripts/verify-changelog-fragment.mjs --base origin/main` | pass, 3 fragments |
| CI (`gh pr checks 82`) | 14 pass, 0 pending; `e2e`/`windows` skipping |

**Mutations re-run independently** (in-memory backup + post-restore assertion, per the durable finding this PR commits — I did not use `git checkout --`). Counts match the PR body exactly:

| Mutation | Claimed red | Measured red |
|---|---|---|
| M5 — service drops `cost` from `judgeAmbiguous` (the #73 survivor) | 3 | **3** |
| M9 — service drops the test-case embedding record (the #77 arm) | 1 | **1** |
| M1 — judge drops its cache-key embedding record | 3 | **3** |
| M6 — `exceeded()` fails closed on unpriced embedding again | 4 | **4** |

**Coverage re-measured** (`--coverage.include='src/lib/testcoverage/**'`), matching the PR table to the decimal: `cost-tracker.ts` 100 / 92.85 · `judge.ts` 100 / 96 · `coverage-service.ts` 97.52 / 80 · `suggestion-generator.ts` 97.08 / 83.78. All clear the 80% line floor.

### Security: CLEAN
OWASP pass over the diff. The two new Prisma filters are literal constants (`"testcoverage.embedding"`, an `in` over two constants) with no user input; no string-built query, no shell, no deserialization, no secret, no new dependency. The values newly persisted as `provider`/`model` come from the embedder backend, not from a request. No auth or scoping surface is touched. One thing I checked and found correct: only three `agentStep` values are ever written under a `testCoverageRun:` session (`cost-tracker.ts:224`, `judge.ts:364`), so today's read-path filter is complete.

### Tests: STRONG
`state.run.embeddingTokens === sum of every batch the embedder was handed` (`coverage-service.test.ts`) is a genuinely load-bearing invariant — it fails if *any* embedder call in the service path goes unbilled, not just the one a given test had in mind. The settable embedder mock with prefix-pinned vectors is a good seam, and pinning the AMBIGUOUS band at cosine 0.70 to make the judge phase actually run is the right way to reach code that was previously untestable from the service.

### Documentation: NEEDS UPDATE
USER_GUIDE is updated and reads well. `docs/ARCHITECTURE.md` is not, and now contradicts the code in two places (comments below). Changelog: three fragments, correct `.changes/unreleased/<issue>-<slug>.md` convention, no `CHANGELOG.md` edit, no version bump.

### Scope
Nothing under `e2e/` or the `e2e` CI job was touched (confirmed against the file list). Read all 14 changed files in full, plus `indexer.ts`, `task-runner.ts`, `routes/test-coverage.ts` and the UI budget tile, which the diff does not touch but which decide whether these findings are real.

### Findings
1. **BLOCKING** — #72 AC1 unmet for the index phase, while the docs claim completeness — `.changes/unreleased/72-in-loop-embedding-usage.md:6`
2. Medium — `ARCHITECTURE.md` now contradicts the code (two sentences) — `server/src/lib/testcoverage/cost-tracker.ts:332`
3. Low — #72 AC2 has no single-test proof — `server/tests/lib/testcoverage/coverage-service.test.ts:650`
4. Low — `readBudget`'s LLM aggregate is a two-value allowlist; a future phase drops out of the total silently — `server/src/lib/testcoverage/cost-tracker.ts:385`
5. Low — the two new view fields have no production consumer and no follow-up issue — `server/src/lib/testcoverage/cost-tracker.ts:74`
6. Informational — one extra `ai_token_usages` insert per judge batch / two per cluster — `server/src/lib/testcoverage/judge.ts:289`
7. nit — `vi.unstubAllEnvs()` in the test body rather than `afterEach` — `server/tests/lib/testcoverage/judge.test.ts:427`

mgcronin:
Round 2 pushed as `af8c0fb` — all 7 findings addressed; see the PR body's **Review round 2** section for the detail.

**The blocking one is fixed at the source, not by narrowing the docs.** `task-runner.ts` now owns one `CoverageCostTracker` for the whole run and threads it into `indexer.index(…, { cost })` and `runCoverageScoring(…, { cost })`, so the `index` phase's case and step embedding batches — the largest embedding consumer a cold run has — are on the budget. The runner flushes it too: with no `caller` wired the service never runs. #72 AC1 now holds for a *run*, which is what it says.

Non-blocking: 2, 3, 4, 5 and 7 taken; 6 declined with a reason (an audit row per embedder call is the feature, and the insert is fire-and-forget). Six new mutations all went red, and M1′ confirms the two new AC2 tests are the ones that catch a dropped in-loop record. Found and filed rather than fixed: the budget endpoint reports zeros for the whole live part of a run, because `flush()` is the only writer of the columns it reads — added to #81.

Still **not merging**; over to you.

mgcronin:
CI green on `8f33ed2` — full re-read, **14 pass, 0 pending**, `e2e` and `windows` skipping (`api` pass 24m38s, `api-outcome` pass).

One thing worth knowing for the next push: the first round-2 push (`af8c0fb`) fired **only** the CLA workflow. GitHub had the PR at `mergeStateStatus: DIRTY` because `main` had moved (#80, #83 landed), and a PR GitHub cannot compute a merge commit for does not run `pull_request`-triggered workflows at all — CI, SAST and CodeQL were silently absent rather than failing. Merging `origin/main` in (`8f33ed2`, clean, no conflicts) brought all of them back. **A check set of one is not a passing check set.**

Re-ran the whole gate after the merge: `pnpm lint` 0, `pnpm typecheck` 0 (8 projects), `pnpm test` 0 — server 17,655 / ui 3,387 / scripts 2,367 / shared 698 / embeddings-svc 147 / copilot-svc 78 / ui-kit 7, no package failed. `verify-changelog-fragment`, `agents:verify`, `skills:verify` all pass.

Not merging.

server/src/lib/testcoverage/cost-tracker.ts
```diff
@@ -312,17 +373,33 @@ export async function readBudget(
...
+  // are read back from its `ai_token_usages` rows (NULL cost). #77 — split by
+  // phase on `agentStep`, so the persisted view answers the same question the
+  // in-memory one does: which share of the unknown spend stops a run.
+  const sessionId = coverageSessionId(runId);
+  // Total first, embedding second, LLM by subtraction — the same `else` that
+  // {@link CoverageCostTracker.view} splits on. An allowlist of the LLM phases
+  // would silently drop a phase added later out of the persisted total.
+  const [unpricedTotal, unpricedEmbedding] = await Promise.all([
+    db.aITokenUsage.aggregate({
+      where: { sessionId, estimatedCostUsd: null },
```
mgcronin, line 385: **Low/Medium (confidence: high — measured) — the scoping key on this new query is not pinned by any test.**

I mutated `where: { sessionId, estimatedCostUsd: null }` to `where: { estimatedCostUsd: null }` and re-ran `tests/lib/testcoverage`: **32 files / 322 tests, all still green**. The shipped code is right, so this is not a live defect — but the guard cannot fire.

It matters more than a generic missing assertion because of what the arm now does. Before this PR, the LLM total came from an `agentStep` allowlist and the test asserted the full `where` object (`toHaveBeenCalledWith({ where: { sessionId: "testCoverageRun:r1", … } })`), so `sessionId` *was* pinned. The refactor to the `usageRows` helper replaced that with a row matcher that filters on `estimatedCostUsd` and `agentStep` only — see the comment on `cost-tracker.test.ts`. The embedding arm keeps a `toHaveBeenCalledWith` including `sessionId` (line ~502); this arm has none. Unscope it and every other run's unpriced LLM tokens land in this run's `unpricedLlmTokens` — the field the budget endpoint uses to say "unknown spend that stops a run".

One line closes it: add `sessionId` to the object the `usageRows` matcher checks, or keep a `toHaveBeenCalledWith` on the total query alongside the existing one.

server/tests/lib/testcoverage/cost-tracker.test.ts
```diff
@@ -405,21 +441,89 @@ describe("readBudget", () => {
...
       breakdown: { embeddingTokens: 10, judgeTokens: 5, suggestionTokens: 2 },
     });
   });
 
+  /**
+   * Answer `aggregate` from a row set instead of by call order, so the test
+   * measures the `where` clauses `readBudget` actually issues rather than
+   * restating them.
+   */
+  function usageRows(
```
mgcronin, line 455: nit / follow-up to the finding on `cost-tracker.ts:385`: the helper is a genuine improvement — answering `aggregate` from a row set measures the `where` clause `readBudget` issues instead of restating call order, and it is what makes the new "a phase it has never heard of" test meaningful (that one goes red under the allowlist mutation, verified).

The gap is only that the matcher ignores `where.sessionId`, so the run-scoping of both aggregates is invisible to it. Threading `sessionId` through the `rows` fixtures and filtering on it would keep the ergonomics and restore the scope assertion the old `toHaveBeenCalledWith` carried.

server/src/lib/testcoverage/task-runner.ts
```diff
@@ -176,6 +189,14 @@ export async function runTestCoverageJob(
       await advance("suggest", "skipped");
     }
 
+    // #72 — with no caller wired the service never runs, so the index phase's
+    // usage would never be persisted. Flushing here covers both paths; the
+    // service's own flush writes the same totals from the same tracker.
+    // Accounting must never be the reason a completed run reads as failed.
+    await cost.flush().catch((err: unknown) => {
```
mgcronin, line 196: **Low (confidence: high) — a run that throws after the index phase never persists what it spent.**

This `flush()` sits inside the `try`, after the suggest phase. The `persist()` writes are fire-and-forget and land independently, so a failed run's `ai_token_usages` rows are there — but `TestCoverageRun.tokenCostCents` / `embeddingTokens` stay at 0, and `readBudget` sources `usedCents` from `tokenCostCents`. An operator looking at a failed run sees `$0.00` spent after real cloud-embedder spend was incurred.

The shape pre-dates this PR (the service's own flush was equally inside its happy path), and this PR does not make the failure more likely — but it does move the largest embedding consumer inside the window, so more real spend is now invisible on that path. A `finally`, or a second flush in the `catch` that marks the run failed, would close it. Not blocking: reporting-only, and the usage rows survive.

Worth folding into #81, which already owns this endpoint's wrong-time reads.

server/src/lib/testcoverage/coverage-service.ts
```diff
@@ -100,10 +112,12 @@ export async function runCoverageScoring(
   const emit = deps.emit ?? noopEmit;
   const embedder = getEmbedder();
 
-  const cost = new CoverageCostTracker(
-    { runId: input.runId, projectId: input.projectId, userId: input.userId },
-    { budgetCents: deps.budgetCents, db },
-  );
+  const cost =
+    deps.cost ??
```
mgcronin, line 116: **Informational — the tracker is now flushed twice on the wired path.**

When `deps.cost` is supplied, `runCoverageScoring` flushes it and then `task-runner.ts:196` flushes the same instance again. `flush()` writes absolute totals through a single `update`, so it is idempotent and the second write is a no-op in effect — one extra `UPDATE` per run, nothing more. Calling it out only so it is a known consequence rather than a surprise later: the doc comment on the runner's flush explains why it is unconditional, and that reasoning is right (with no caller wired, nothing else would persist the index phase).

The `deps.cost ?? new CoverageCostTracker(...)` seam itself is good — the service still stands alone, and the `budgetCents`-is-unused-when-`cost`-is-supplied caveat is documented at the interface.

server/src/lib/testcoverage/indexer.ts
```diff
@@ -100,16 +126,35 @@ export class TestCoverageIndexer {
     // Embed all case texts and step texts in one batch each.
     const caseTexts = todo.map((c) => caseText(c.case));
     const caseEmb = await embedder.embed(caseTexts);
+    // #58 rule, applied to the index phase: the embedder key is read AFTER
+    // `embed()` (a failed backend may have been swapped for the hash stub) and
+    // the model is the one `embed()` itself reported.
+    options.cost?.record({
```
mgcronin, line 132: **Informational — a behaviour change worth a sentence somewhere.**

Index-phase spend now counts toward `usedCents`, and `exceeded()` is consulted before each judge batch. On a priced cloud embedder with a large cold corpus, the index phase can therefore eat into the 20¢ cap before any judging happens, and in the limit a run could reach the cap during indexing and skip the LLM phases entirely. That is the budget doing exactly what #72 asks of it, so it is not a defect — but the changelog says "reaches the run budget" without saying "and can therefore consume it", and an operator who sees a run stop before judging will want that sentence.

The billing itself is right: `embedder.key` read *after* `embed()` and the model `embed()` reported (the #58 rule), both batches billed, nothing billed when every case is already indexed. `EmbeddingCostRecorder` being a narrow bill-only interface rather than the whole tracker is a nice touch — the indexer structurally cannot consult the budget it has no business consulting.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking findings, 4 non-blocking.

Automated analysis, not a human sign-off: the authenticated `gh` user is also this PR's author, so GitHub cannot record an independent approval from it. The merge decision stays with the caller.

## Cycle-1 blocking finding: fixed, and verified by running it

The cycle-1 blocker was that `task-runner.ts` runs the `index` phase — the run's largest embedding consumer — before `runCoverageScoring` constructs a tracker, so #72 AC1 ("every embedder call a **run** makes") was false while the changelog and USER_GUIDE told an operator it was true. The author declined the cheap remedy (narrow the docs) and wired the phase instead. I re-verified that by mutation, not by reading the response — each mutation applied alone, restored from an out-of-tree backup, and the restore asserted with `git status --porcelain` (clean at `8f33ed2` after every one):

| Mutation | Result |
|---|---|
| Runner withholds `{ cost }` from `indexer.index` | **3 red** (incl. both new #72 runner tests) |
| Indexer stops billing the case batch | **2 red** |
| Runner never flushes | **1 red** |
| Service ignores the injected `deps.cost` | **1 red** |
| `readBudget` back to the two-value LLM allowlist | **1 red** |
| Judge drops its cache-key embedding billing | **5 red**, including *both* new AC2 tests — they are not vacuous |

I also checked the claim itself rather than the tests: `grep -n "\.embed("` over `server/src/lib/testcoverage/` finds exactly seven embedder calls (indexer x2, coverage-service x2, judge x1, suggestion-generator x2) and every one has an adjacent `cost.record`. The `import` phase makes none. AC1 now holds for the whole run, and `docs/ARCHITECTURE.md`'s two contradicting passages are corrected.

## Non-blocking findings (4)

See the inline comments. The most substantive is a **scoping key that no test pins**: removing `sessionId` from the new unpriced-total aggregate leaves all 322 tests green. The shipped code is correct — this is a guard that cannot fire, not a live defect.

## Verified, not taken on trust

- Gates re-run in the PR worktree: `pnpm typecheck` 0 (8 projects); `eslint` 0 on the changed files; `npx vitest run tests/lib/testcoverage tests/lib/finops` = **32 files / 322 tests passed**.
- Coverage re-measured on the touched files (80% floor): `cost-tracker.ts` 100/92.85 · `judge.ts` 100/96 · `indexer.ts` 98.07/100 · `coverage-service.ts` 97.52/80.51 · `suggestion-generator.ts` 97.08/83.78 · `task-runner.ts` 95.12/85.71 (lines/branches). All above the floor; matches the PR table.
- CI on `8f33ed2` re-read plainly: **14 pass, 0 pending**, `e2e` and `windows` skipping; `api-outcome` pass; `mergeStateStatus` CLEAN. The cycle-1 "a check set of one is not a passing check set" problem is resolved by the `origin/main` merge.
- `pnpm agents:verify`, `pnpm skills:verify` and `scripts/verify-changelog-fragment.mjs` all pass; three fragments present under `.changes/unreleased/`, `CHANGELOG.md` untouched, no version bumped.
- Security: no dependency added, no new user input reaches a query, no secrets, no auth or tenancy surface touched. The one new query shape (`where: { sessionId, estimatedCostUsd: null }`) is Prisma-parameterised with a derived constant.
- Scope fence respected: nothing under `e2e/`, no change to the `e2e` CI job.
- Durable finding from cycle 1 is committed in-tree at `.claude/agent-memory/code-issue/` and indexed; `MEMORY.md` is 1,356 B, well inside the 17,500 B budget.

Good work on the two AC2 tests in particular: pinning the published per-token price on all six rows *and* using a `MODEL_PRICES` override large enough that cents can move is the right answer to "fractions of a cent cannot show it", and identifying the in-loop rows by set-difference on token counts rather than row order makes them robust to reordering.

Issue #99, opened 2026-09-23
scripts tests: the two whole-tree gate arms time out under the suite's own parallelism (4.1s of work in a 5s budget)


mgcronin:
Found while running the gate for #85/#86. Not caused by that change — it touches no file under `scripts/`, and the two tests scan **tracked** files, so an unstaged branch cannot affect them.

### Measured

`pnpm test` (and `pnpm --filter @metis/scripts test` on its own) fails with two timeouts:

```
❯ lib/verify-no-company-identifiers-runner.test.mjs (18 tests | 1 failed) 17042ms
    × passes the gate with whatever list this environment has 8064ms
❯ lib/gate-cwd-anchoring.test.mjs (9 tests | 1 failed) 19824ms
    × produces identical output from the repository root and from server/src 16423ms
Error: Test timed out in 5000ms.
```

Run those two files **alone** and both pass — 27 tests, no failures. The gate they spawn is genuinely slow on its own:

```
$ time node scripts/verify-no-company-identifiers.mjs
verify-no-company-identifiers: 4078 tracked text files and 4082 paths scanned … none found.
real    0m4.110s
```

4.1 s of real work against vitest's default 5 s `testTimeout`, spawned as a subprocess, with 57 other test files competing for the same machine. `gate-cwd-anchoring` runs it **twice** (root and `server/src`) inside one 5 s budget, so it cannot pass under any load at all. The margin shrinks every time a file is added to the repo.

This is the #1379 shape: a file line whose duration ≈ the package's `testTimeout` is a timeout, not a failure — and it is a signal-quality problem, because a red `scripts` job that means "the machine was busy" is indistinguishable at a glance from one that means a private term was reintroduced.

Per #1379 the answer is to cut the test's tie to the contended resource, not to raise the timeout: a bigger number just moves the flake to the next tree that is 20% larger.

### Acceptance criteria

- [ ] `pnpm --filter @metis/scripts test` passes with the full suite running in parallel, repeatably.
- [ ] The whole-tree acceptance arms no longer spend their budget re-scanning 4,000 files per assertion — e.g. spawn the gate once per file and assert several things about the one result, or run the two whole-tree arms in a pool that is not competing with the other 57 files. Raising `testTimeout` alone does not close this.
- [ ] Whatever the fix, the arms still fail when a private term IS present in the tree — that acceptance property is the reason they exist.

Issue #98, opened 2026-09-23
Indexing errorMessage still reaches the UI verbatim — raw exception text on the documentation status line


mgcronin:
Found while implementing #86 (the generation `errorMessage` arm of the same exposure). Filed rather than folded into that PR because it is a different column, a different pipeline and needs its own vocabulary — `generationFailureMessage` classifies *document-generation* failures and would be the wrong classifier here.

### The path

1. **Write** — `server/src/lib/rag/ingest-queue.ts:192`, `:204`, `:219` and `server/src/lib/rag/knowledge-service.ts:2023` put `lastError.message` / `message` straight into `Document.errorMessage`. That is the exception's own text: embedder HTTP bodies, LanceDB errors, absolute paths, provider payloads.
2. **Read** — `server/src/routes/generated-docs.ts:421` (and the list handler's `indexByGeneratedDocId`, `:338`) return `syntheticDocument.errorMessage` verbatim as `indexing.errorMessage`. Unlike the sibling field two lines up, it is **not** routed through `publicGenerationErrorMessage`.
3. **Render** — `ui/src/app/(authed)/projects/[id]/documentation/page.tsx:1099` and `:1101`:

   ```ts
   case "rejected":
     return indexing.errorMessage?.trim() || "Indexing was rejected.";
   default:
     return indexing.errorMessage?.trim() || "Queued for indexing.";
   ```

So the exact exposure #52 closed for a failed generation, #67 closed for the `warnings` column and #86 closed for the legacy `errorMessage` blob is still open on the indexing status line, for **new** rows as well as old ones.

`knowledge-service.ts:490` (`errorMessage: parsed.reason`) and the `"deleted"` writes are METIS-authored and fine; the three `ingest-queue` writes and `knowledge-service.ts:2023` are the ones carrying `String(err)`.

### Acceptance criteria

- [ ] A fixed vocabulary for indexing failures (the shape of `generationFailureMessage`, not a reuse of it): quarantine/rejection reasons, embedder unavailable, provider auth, rate limit, generic. The raw error stays in the server log.
- [ ] Every read path that serves `Document.errorMessage` to a client routes through it — the detail handler, the list handler, and the socket `document:status` payload (`ingest-queue.ts:192`) which carries `lastError.message` to the browser directly.
- [ ] A test that fails before the fix, using a row whose `errorMessage` holds a provider response body with a path and an API-key-shaped string in it.
- [ ] Old rows are covered on read, as #86 did: the column is already full of raw text.

PR #100, opened 2026-09-23
fix: Resolve #85, #86 — AggregateError sub-errors are logged; the legacy errorMessage column stops echoing exceptions

### LINKED ISSUE #85: Error serialisation follow-ups from #68: AggregateError.errors dropped, two other redaction sinks still rebuild
Follow-ups from the review of #80. Both non-blocking there.

### 1. `AggregateError.errors` is still lost (Low)
`server/src/lib/logger.ts:175` (`errorShape`): `AggregateError.errors` is an **own but non-enumerable** property (verified), so it is dropped by both the explicit key list and the `...err` spread — the same `{}` defect #68 fixed, one class down. Any `Promise.any` / batched-connector rejection logs an `AggregateError` with no sub-errors.

### 2. Two other redaction sinks still rebuild from `Object.entries` (Low / informational)
- `server/src/lib/audit/audit-service.ts:194`
- `server/src/lib/sandbox/audit/redact.ts:179`

Both **persist** what they serialise, so writing stacks there is a product decision (retention, PII) rather than an obvious fix — hence a separate issue rather than part of #68.

### Acceptance criteria
- [ ] `AggregateError` logged in metadata serialises `name`, `message`, `stack` **and** each entry of `errors`, with a test that fails before the fix.
- [ ] A decision recorded (code comment or ADR) on whether the two audit sinks should serialise Errors; if yes, implemented with tests; if no, the reason stated where the next reader will find it.

### LINKED ISSUE #86: Legacy errorMessage field still surfaces raw exception text in the UI (follow-up to #67)
Follow-up from the review of #80 (Medium, non-blocking there because it predates the PR).

#67 closed the raw-exception exposure in the `warnings` column. The **legacy `errorMessage` column is still exposed**:

- `server/src/lib/docs-gen/generation-failure-message.ts:118` — `publicGenerationErrorMessage` returns a non-`failed` row errorMessage **verbatim**.
- Pre-#252 degraded rows stored `DocWarning[]` JSON in that column, built by the *old* `sectionFailedWarning(label, String(err))` — i.e. with the exception string embedded.
- The UI still parses it: `ui/src/app/(authed)/projects/[id]/documentation/page.tsx:1121`, `resolveDocWarnings` legacy fallback.

So the exact exposure #67 closed survives for any document generated before #252. The in-code justification is that the blob is "METIS-authored, so safe" — METIS *formatted* it, but it embedded the exception.

### Acceptance criteria
- [ ] Legacy `errorMessage` content reaching the client is run through the fixed failure-message vocabulary, or the legacy fallback parse is removed and those rows render a fixed message.
- [ ] A test that fails before the fix, using a pre-#252-shaped `errorMessage` row with an exception string in it.
- [ ] No raw exception string, stack or provider payload reaches any user-visible field for old or new rows.

mgcronin:
Closes #85
Closes #86

Both are follow-ups from the review of #80, and both are the same defect family one arm further on: something own-but-non-enumerable being dropped by an `Object.entries` rebuild (#85), and a legacy column that was exempted from a sanitiser on a reason that did not hold (#86).

## #85 — `AggregateError.errors` was dropped

`errors` has the identical property descriptor to the `name` / `message` / `stack` / `cause` that #68 had to enumerate explicitly: own, **not** enumerable. It was on neither the explicit key list nor reachable through the `...err` spread, so `errorShape` (`server/src/lib/logger.ts`) rebuilt every aggregate without its sub-errors. Every `Promise.any` and every batched-connector rejection logged

```json
{ "name": "AggregateError", "message": "All promises were rejected" }
```

— which is exactly why nobody noticed: the line looks complete, and the part that says *why each attempt failed* is the part that is missing.

The fix captures `errors` when it is an array. It is duck-typed rather than `instanceof AggregateError` so a cross-realm aggregate is covered; the caller already recurses into the result, so every sub-error comes back through `errorShape` and is redacted by the same key rules. A credential hung off a sub-error still reads `[REDACTED]` — asserted, because a fix that bypassed the recursive walk would satisfy the "sub-errors survive" assertion on its own.

## #85 part 2 — the decision on the two persisting sinks

`server/src/lib/audit/audit-service.ts` and `server/src/lib/sandbox/audit/redact.ts` carry the same rebuild. The issue asked for a decision, and the decision is **no** — recorded in `docs/decisions/0016-error-serialisation-in-the-persisting-sinks.md` with three reasons, the first of which is evidence rather than taste:

1. **Nothing reaches them.** Every `audit({...})` call site in `server/src` already reduces its error at the boundary (`(err as Error).message`, `toConnectorError(err).code`, `error.slice(0, 200)`), and so does the sandbox emitter's own failure path. That is a claim about the source, so a new block in `redaction-sinks.enumeration.test.ts` reads it off the source and fails if a bare error ever appears in one of those calls.
2. A stack in an exported `AuditLog` row carries absolute server paths, and a `cause` chain can drag a whole provider payload into a retained compliance record. The logger's transport is rotated and operator-facing, which is precisely why the same content is right there and wrong here.
3. The call site knows what the row is for. Widening the sink would let a future caller dump an exception into a compliance row by accident instead of by decision.

Made executable rather than left as prose:

- all three sinks declare `ERROR_SERIALISATION_POLICY:` in-file; the registry test requires a recognised marker on **both** axes and fails if the two registries name different files;
- `audit-redaction.test.ts` and `lib/sandbox/audit/redact.test.ts` each assert that an Error yields no `name` / `message` / `stack` / `cause` / `errors` **and** that a credential hung off it still redacts — so copying the #68 repair into either file by analogy goes red and names the ADR.

ADR 0008's own warning ("changing a sink's policy with no observed defect is how the next fail-open gate gets written") is what ruled out the tempting middle answer of "serialise `name` and `message` but not `stack`".

## #86 — the legacy `errorMessage` column

`publicGenerationErrorMessage` returned a non-`failed` row's message **verbatim**, on the reasoning that a legacy `degraded` row's blob is METIS-authored. METIS *formatted* it; the pre-#67 `sectionFailedWarning(label, String(err))` embedded the exception inside it, and the UI's `resolveDocWarnings` legacy fallback parses and renders exactly that. So the exposure #52 closed for a failed row and #67 closed for the `warnings` column survived intact for every document generated before #252.

The blob is now re-derived instead of trusted: a `DocWarning[]` goes through `publicDocWarnings` — the same classifier the `warnings` column gets — and anything that is not a warning list (a raw exception string a pre-#52 row left behind, a JSON object, `"null"`) collapses to the fixed message.

Two things had to hold together, and both are asserted: nothing is echoed, **and** the sanitised blob still parses into the `DocWarning[]` shape the UI's fallback requires (it drops any entry without a string `message`), so sanitising it does not silently blank the banner it was meant to make safe. A recognisable provider failure stays recognisable — a 402 embedded in a legacy blob still comes out as the balance message.

The UI parse is kept rather than removed, because it is what renders legacy warnings at all; the comment at `resolveDocWarnings` now records that the input is sanitised **upstream**, not there.

- **The #67 enumeration gate could not tell prose from code.** `docs-gen-warning-detail.enumeration.test.ts` scanned RAW source, so a doc comment that merely *named* the old `sectionFailedWarning(label, String(err))` shape was reported as a call site passing an exception — and #86's fix, whose whole subject is that builder, tripped the gate it was honouring. A gate that makes documenting the defect it guards against impossible gets the documentation reworded instead of the code fixed. `blankComments` (which the sibling redaction scanner already had, for the identical reason) moves to `server/tests/helpers/source-scan.ts`; the scanner is split so the gate itself can be run against a fixture, and a new arm asserts that offending text in comments yields no call site **while the one real call in the same fixture is still found**.
- **#98** — the indexing `errorMessage` is the same exposure on a different column: `ingest-queue.ts` writes `lastError.message` into `Document.errorMessage`, `generated-docs.ts:421` returns it as `indexing.errorMessage` without routing it through anything, and the UI renders it at `page.tsx:1099`. For new rows as well as old. Filed rather than folded in: it needs its own vocabulary, since `generationFailureMessage` classifies *generation* failures.
- **#99** — two `scripts` tests time out under the suite's own parallelism (4.1 s of whole-tree scanning against vitest's 5 s default; `gate-cwd-anchoring` does it twice in one budget). Not caused by this branch — it touches no file under `scripts/`, and those tests scan *tracked* files. They pass when run alone, which is how the diagnosis was made.

## Tests — the mutations that proved each one can fail

Every assertion block below was watched going red, then restored.

| # | Mutation | What went red |
|---|---|---|
| 1 | `if (Array.isArray(nested)) out.errors = nested` → `out.errors = nested ?? []` | `leaves a non-array errors property alone` (1 test) |
| 2 | the `errors` capture deleted — the #85 defect restored | 4 of the 5 `AggregateError sub-errors (#85)` tests |
| 3 | `if (status !== "failed") return errorMessage` — the #86 defect restored | 5 `legacy errorMessage warning JSON (#86)` tests |
| 4 | the array branch returns the blob unchanged instead of re-deriving it | 3 of those 5, including `does not echo the exception` |
| 5 | the #68 repair copied into `audit-service.ts`, its marker renamed | 2 `audit sink — Errors are reduced at the call site` tests + the marker gate |
| 6 | `metadata: { cause: err }` planted at a real audit call site (`jira-service.ts:284`) | `no audit call site passes a bare error` |
| 7 | `sandbox/audit/redact.ts` dropped from `REGISTERED_ERROR_POLICIES` | `registers the same set of files on both axes` |
| 8 | the #68 repair copied into `redactSandboxPayload` | `does not persist name, message, stack, cause or sub-errors` |
| 9 | `blankComments` removed from the #67 scanner | `does not mistake prose…` **and** `no call site passes an exception's own message` |
| 10 | blanking made over-greedy (blanks the identifier itself) | `does not mistake prose…` **and** `finds the call sites at all` |

The premise scan also carries its own positive control: three planted call-site shapes must be detected and three reduced shapes must not be, so deleting a pattern cannot read as "the premise holds".

## Coverage

On the four touched source files (`logger.ts`, `docs-gen/generation-failure-message.ts`, `audit/audit-service.ts`, `sandbox/audit/redact.ts`), measured with the same command before and after:

| | Stmts | Branch | Funcs | Lines |
|---|---|---|---|---|
| before | 96.25% | 89.15% | 95.12% | 96.15% |
| after | **96.42%** | **89.41%** | **95.23%** | **96.31%** |

`generation-failure-message.ts` is 100% statements and lines. Floor is 80%.

## Gate

- `pnpm lint` — clean (4,082 tracked files scanned, eslint `--max-warnings=0`).
- `pnpm typecheck` — clean, all 8 projects.
- `pnpm --filter @metis/server test` — 17,681 passed, 11 skipped, 0 failed (1,127 files).
- `pnpm --filter @metis/ui test` — 3,387 passed, 1 skipped. `@metis/shared` 698 passed, `@metis/ui-kit` 7 passed.
- `scripts` — 2 timeouts, environmental and pre-existing; filed as #99 with the isolation evidence.

## Security

No dependency change, so nothing to audit. OWASP pass over the diff: #86 *closes* an information-disclosure path (A01/A09) and opens none — the new `JSON.parse` is wrapped, and `publicDocWarnings` builds fresh objects by spread, so a `__proto__` key in a legacy blob becomes an own data property rather than polluting anything. #85 widens what is logged, not what is served, and every sub-error is walked by the same key redaction as the aggregate. No secrets, no SQL, no shell, no path handling.

No version bumped. One changelog fragment per issue, matching this repo's precedent for multi-issue PRs.


server/src/lib/docs-gen/generation-failure-message.ts
```diff
@@ -111,20 +111,54 @@ export function generationFailureMessage(err: unknown): string {
...
+ */
+function legacyWarningJson(errorMessage: string): string {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(errorMessage);
+  } catch {
+    return generationFailureMessage(errorMessage);
+  }
+  if (!Array.isArray(parsed)) return generationFailureMessage(errorMessage);
+  return JSON.stringify(publicDocWarnings(parsed));
```
mgcronin, line 159: **Low / high confidence — the sanitiser is element-wise, but the comment and the changelog say it is whole-blob.**

`legacyWarningJson` decides on array-ness alone and then delegates per entry, and `publicDocWarnings` only re-derives `message` for entries whose `kind === "section-failed"` and that lack `detailSafe`. Everything else in an untrusted legacy blob is echoed verbatim. I confirmed four shapes by probe against this branch:

| legacy blob | result |
|---|---|
| `["Error: connect ECONNREFUSED /srv/metis/secrets.env"]` | echoed unchanged |
| `[{kind:"source-unavailable", message:"boom: /srv/metis/x.ts SELECT *"}]` | echoed unchanged |
| `[{kind:"section-failed", …, detail:"/srv/metis/x.ts SELECT *"}]` | `message` re-derived, `detail` echoed (`{...w}` keeps unknown keys) |
| `[{kind:"section-failed", section:"/srv/metis/x.ts", …}]` | `section` copied verbatim **into** the new message |

None of these is a regression — it is exactly the contract `publicDocWarnings` already has for the `warnings` column since #67, and the known pre-#252 writer only embedded `String(err)` in a `section-failed` `message`. So this is not blocking. But the docstring above ("anything that is not a warning list … collapses to the fixed message"), the ADR, and the changelog fragment all read as though the whole blob is re-derived, and the next reader will believe them. Either soften those three sentences to say *per-entry, `section-failed` only, other fields passed through*, or tighten the legacy arm to rebuild only known fields (`kind`, `section`, `message`, `severity`) and collapse an entry that is not a plain object. The second is cheap here precisely because the legacy column has one known writer.

server/src/lib/logger.ts
```diff
@@ -176,6 +184,17 @@ function errorShape(err: Error): Record<string, unknown> {
...
+  // NOT enumerable. Omitting it from this list dropped every sub-error of a
+  // `Promise.any` / batched-connector rejection, leaving only the aggregate's
+  // own "All promises were rejected" — which is why nothing looked broken.
+  // Duck-typed on an array rather than `instanceof AggregateError` so a
+  // cross-realm aggregate, or a library error that copies the shape, is covered
+  // too; a non-array `errors` is left to the spread below, as any other own
+  // property is. The caller recurses into the result, so each entry that is
+  // itself an Error comes back through this function and is redacted by key.
+  const nested = (err as { errors?: unknown }).errors;
+  if (Array.isArray(nested)) out.errors = nested;
```
mgcronin, line 197: Good fix, and duck-typing on the array rather than `instanceof AggregateError` is the right call for the cross-realm case — the test that a non-array `errors` is left to the spread, and that a plain `Error` gains no empty `errors` key, is what makes that safe.

**nit (Low):** the capture is unbounded in *count*. `redact`'s `depth > 6` cap bounds nesting, not width, so a `Promise.any` fan-out over N connectors now writes N sub-error stacks into one log line. That is the point of the fix for a handful of connectors; for a wide batched fan-out it is a log-volume change nothing bounds. If any call site can aggregate more than a few dozen, consider `nested.slice(0, N)` with a `truncated` count. Not blocking — I found no call site in `server/src` that fans out that wide.

ui/src/app/(authed)/projects/[id]/documentation/page.tsx
```diff
@@ -1107,6 +1107,17 @@ function formatIndexingSummary(indexing?: GeneratedDoc["indexing"]): string {
...
+ *
+ * #86 — the legacy fallback is kept, and it is safe because the SERVER sanitises
+ * that column now: `publicGenerationErrorMessage` re-derives every
+ * `section-failed` warning in a pre-#252 blob through the fixed
+ * `generationFailureMessage` vocabulary, and collapses anything that is not a
+ * warning list (a raw exception string a pre-#52 row left behind) to a fixed
+ * message. Before that fix this parse rendered `String(err)` — provider
+ * response bodies, absolute paths, SQL text — verbatim in the banner, which is
+ * exactly the exposure #67 closed for the `warnings` column. Do not start
+ * trusting this input on the strength of that: it is sanitised upstream, not
```
Comment (mgcronin) [inline ui/src/app/(authed)/projects/[id]/documentation/page.tsx:1119]: The comment is accurate and the decision to keep the legacy parse is right (removing it would blank legacy banners entirely rather than make them safe).

**Low — requirement residual, already disclosed:** #86's third criterion says no raw exception reaches *any* user-visible field. `formatIndexingSummary`, ~20 lines above this, renders `indexing.errorMessage`, which `server/src/routes/generated-docs.ts:338` and `:421` return straight off `Document.errorMessage` (written from `lastError.message` by the ingest queue) for new rows as well as old. You filed that as #98 and said so in the PR body, which is the right handling — but this PR says `Closes #86`, so please note in the closing comment that criterion 3 is met for the generation column only and #98 carries the indexing column. Otherwise the criterion reads as satisfied by a closed issue.

server/src/lib/audit/audit-service.ts
```diff
@@ -17,6 +17,20 @@
  * is exactly the accounting an audit record exists to carry. Rationale and the
  * other two sinks' policies: `docs/decisions/0008-redaction-sinks.md`.
  *
+ * ERROR_SERIALISATION_POLICY: reduce-at-call-site
```
mgcronin, line 20: The ADR reaches the right answer and, more usefully, makes it executable rather than leaving prose: the two-axis registry, the symmetry arm, the behavioural assertions in both sink tests, and a premise scan with a positive control. Declining to copy the #68 repair into a retained, exported compliance row is well argued, and ADR 0008's own warning is the right authority to cite.

**nit (advisory):** `errorPolicyOf` (and `policyOf` before it) takes the *first* `ERROR_SERIALISATION_POLICY:` match in raw source, comments included — which is necessary, since the marker lives in a comment. It means a file that quotes another sink's marker string above its own would be read as declaring the wrong policy. It is fine today: `logger.ts` discusses the persisting sinks' policy in prose without the marker token. Worth an anchored match (e.g. require the marker at the start of a ` * ` line in the module header) if a third sink ever lands.

mgcronin (commented):
Recommendation: merge with follow-ups — 0 blocking findings, 4 non-blocking.

This is automated analysis, not a human sign-off: the reviewing identity is the PR author, so GitHub cannot record an independent approval here. The merge decision stays with a human.

### Requirements — PASS (with one documented residual)
**#85** — both criteria met. `AggregateError.errors` is captured in `errorShape` (`server/src/lib/logger.ts:196-197`) and each entry is walked back through the same recursive key redaction; the decision on the two persisting sinks is recorded in `docs/decisions/0016-…` **and** made executable (a per-sink `ERROR_SERIALISATION_POLICY:` marker, a two-axis registry gate with a symmetry arm, behavioural assertions in both sink tests, and a source scan of the premise with a positive control).

**#86** — criteria 1 and 2 met on the generation `errorMessage` column. Criterion 3 ("no raw exception string … reaches **any** user-visible field") is met for that column only: `indexing.errorMessage` (`Document.errorMessage`) is still returned unrouted at `server/src/routes/generated-docs.ts:338` and `:421` and rendered by `formatIndexingSummary`. The author found this, filed **#98**, and said so — the right call, but #86 should not be closed silently against criterion 3; say in the closing comment that #98 carries the rest.

### Falsifiability — verified independently, not taken on trust
I reverted three production changes in the branch worktree and watched the suite:
- `errors` capture deleted (the #85 defect restored) → **4 tests red** in `logger.test.ts`, as claimed;
- `if (status !== "failed") return errorMessage` (the #86 defect restored) → **5 tests red**, as claimed;
- `blankComments` removed from the #67 scanner → **2 red**, including the new prose-vs-code arm, as claimed.
Baseline on the 6 touched test files: 147 passed / 0 failed. All mutations restored; worktree left clean.

### Security — no new exposure, one path narrowed
#86 closes an information-disclosure path (A01/A09) on the legacy column. The new `JSON.parse` is wrapped; `publicDocWarnings` rebuilds by spread, so a `__proto__` key in a legacy blob lands as an own data property rather than polluting a prototype — I checked that claim, it holds. #85 widens what is **logged**, not what is served, and sub-errors go through the same key redaction (asserted both ways in the tests). No dependency change, no SQL, no shell, no path handling.

### Tests / coverage
Test quality is above the bar for this repo: both directions asserted on the same run (sub-errors survive **and** a credential on a sub-error still redacts), anti-vacuity arms on both new source scans, and a positive control so deleting a pattern cannot read as "the premise holds". I did not re-measure coverage; the touched-file numbers reported (96.4% stmts) are well clear of the 80% floor and CI's `api` job is green.

### CI
All 16 checks present, zero pending, on a plain re-read: 14 `pass`, `e2e` and `windows` `skipping`. `api` genuinely passed (23m11s) and `api-outcome` passes, so this is not the supersede shape.

**Process note, not a code finding:** this branch is based on `dbedcea`, which predates **#84** (merged 00:58 UTC) — the PR that makes `e2e` run on pull requests. That is why `e2e` reads `skipping` here. Updating the branch from `main` before merge would let the newly-gating job actually run against this change; the risk is low (the only UI edit is a comment), but the check currently proves nothing.

### Credit
The #67 enumeration gate defect found along the way — a source scanner that could not tell prose from code, so documenting the very shape it guards against tripped it — is the kind of thing that normally gets the *prose* reworded instead of the gate fixed. Extracting `blankComments` to a shared helper and asserting both arms (prose yields nothing, the real call is still found) is the right repair.

### Scope of this review
Read the full diff and all four touched source files in full, plus `generated-docs.ts`, `degraded-warnings.ts` and `resolveDocWarnings` in the UI; ran the six touched test files and three mutations in the branch worktree; verified CI, the changelog-fragment convention, the ADR numbering and that no protected path (`e2e/`, `.github/`, `server/dev.db`, root `.env`) is touched. I did not run the full monorepo gate and did not reproduce the two `scripts` timeouts filed as #99 — the PR touches no file under `scripts/`, and CI is green, so I accept that claim as pre-existing rather than verified.

PR #101, opened 2026-09-23
fix: Resolve #88, #94 — close the impact-analysis detail access hatch, scope the drift reads, and re-pin the cost-tracker aggregate

### LINKED ISSUE #88: A01: impact-analysis detail and /sync/drift routes do not check project access
Two pre-existing A01 (broken access control) gaps surfaced by the review of #83. Neither was widened by that PR — it is the first UI consumer of one of them, which is why they are worth closing now.

### 1. Impact-analysis detail path keeps the "empty projectIds = visible to everyone" escape hatch
#70 removed that arm from the *list* path. The twin survives on the detail path:
- `server/src/routes/impact-analysis.ts:431` — `detail.projectIds.length > 0`
- `server/src/lib/impact-analysis/impact-analysis-read.ts:957` — item-only derivation

For an in-flight run (no items yet) any `analysis.read` holder can read it by id, and likewise through the export and Jira-publish routes. Discovery is narrower than the list path was, but it is still a read of another project's analysis. #83 persists `impact_analysis_projects`, so the data needed to close this now exists.

### 2. `GET /sync/drift` and `GET /sync/drift/count` never check project access
- `server/src/routes/sync.ts:167` — returns drift content
- `server/src/routes/sync.ts:189` — returns the count

Both check only the `sync.read` permission, never whether the actor can access the project whose drift they are asking for. `ui/src/hooks/use-drift-count.ts` (added in #83) is their first UI consumer.

### 3. Low, same area: a legacy run can now be invisible to the person who started it
With the list hatch removed, a pre-migration run with no persisted projects and no items is visible to nobody, because `startedById` is not consulted: `server/src/lib/impact-analysis/impact-analysis-read.ts:888-899`.

### Acceptance criteria
- [ ] Detail, export and Jira-publish paths for an impact analysis check project access against the persisted `impact_analysis_projects` rows, with no empty-list escape hatch.
- [ ] Both `/sync/drift` routes check project access in addition to `sync.read`.
- [ ] A legacy run with neither persisted projects nor items remains visible to its `startedById` actor.
- [ ] A test per item that fails before the fix, asserting a cross-project actor is refused.

### LINKED ISSUE #94: Unpriced-total aggregate loses its sessionId scope with all 322 tests still green
From the cycle-2 review of #82 (non-blocking there; the code is correct, the guard is not).

### 1. The unpriced-total aggregate is unscoped in effect
`server/src/lib/testcoverage/cost-tracker.ts:385` — dropping `sessionId` from the new unpriced-total aggregate leaves **all 322 tests green**. Without that scope, every other run's unpriced LLM tokens would land in this run's `unpricedLlmTokens`.

The cause is a test refactor, not the feature: `server/tests/lib/testcoverage/cost-tracker.test.ts:455` moved to a `usageRows` row-matcher and **removed** the `toHaveBeenCalledWith` that previously pinned `sessionId` on this arm. The embedding arm still has one, which is why only this arm is exposed.

This is the repo's recurring "gate that cannot fail" shape: the assertion exists, but nothing makes it fire.

### 2. Behaviour change not mentioned in the changelog
`server/src/lib/testcoverage/indexer.ts:132` — index-phase spend now counts toward `exceeded()`, so a large cold corpus on a priced cloud embedder can consume the cap before judging starts. Intended, but an operator reading the shipped fragment would not know.

### Acceptance criteria
- [ ] Removing `sessionId` from the `cost-tracker.ts:385` aggregate turns at least one test red (state which, and show it red before the fix).
- [ ] Every scoping key the replaced `toHaveBeenCalledWith` pinned is asserted again on both arms.
- [ ] The docs note that index-phase embedding spend counts against the run cap.

mgcronin:
Closes #88
Closes #94

Two pre-existing defects, bundled because both are "a guard that was present but could not fire".

### 1. The impact-analysis detail path kept the empty-`projectIds` escape hatch

`#70` removed the `projectIds.length === 0` arm from the **list** path. Its twin survived on the **detail** path, and the reason it survived is one line in the read layer:

```ts
// server/src/lib/impact-analysis/impact-analysis-read.ts (before)
const projectIds = [...new Set(items.map((i) => i.projectId))];
```

The detail projection derived a run's projects from its `ImpactItem` rows **alone**. A run that has not written an item yet — every pending/running run, every run that failed early or extracted zero changes — therefore reported `projectIds: []`, and the route guard read that as "this run names no project, skip the check":

```ts
// server/src/routes/impact-analysis.ts (before)
if (!isAdminActor(actor) && detail.projectIds.length > 0) { … }
```

Any holder of `analysis.read` could read another project's in-flight run by id. `loadAccessibleImpactDetail` is shared by the **detail**, **markdown export**, **re-run**, **drift** and **Jira-publish** routes, so all five admitted it.

**Fix.** `getImpactAnalysisDetail` now loads the run's persisted `impact_analysis_projects` selection (written with the run since #70) and unions it — started-for first — with the item-derived ids, exactly as the list projection does. The route's empty-list arm no longer means "skip"; it means "unattributable", handled below.

### 2. A legacy run must stay visible to its starter

A pre-#70 run with neither persisted projects nor items is genuinely unattributable. Making the check unconditional would hide it from everyone, including the person who started it. `ImpactAnalysisDetail` now carries `startedById`, and both paths admit exactly that one non-admin principal (plus admins) for a run with no recoverable projects. A run whose projects **are** known is still judged on those alone — starting a run never buys back access to a project you have lost.

### 3. `GET /sync/drift` and `GET /sync/drift/count` never checked project access

Both gated on the `sync.read` **role**, which says the caller may read drift — not *whose*. The project id arrives in the query string, so both now pass the canonical object-level seam `assertProjectAccess` (the same predicate `requireProjectAccess` applies to `/projects/:projectId/*`). A non-member gets **404, never 403**, so probing cannot enumerate projects. The first consumer, `ui/src/hooks/use-drift-count.ts`, already fails silently (`retry: false`, `?? 0`), so a refused badge degrades to `0` with no error surface — no UI change was needed.

## #94 — the unpriced-total aggregate lost its `sessionId` with all tests green

Dropping `sessionId` from the unpriced-**total** aggregate at `cost-tracker.ts:385` left every test passing. The code was correct; the guard was not. Two things had to be true at once:

- the row matcher introduced at `cost-tracker.test.ts:455` filtered on `estimatedCostUsd` and `agentStep` and **ignored `sessionId` entirely**, and
- the same refactor **removed** the `toHaveBeenCalledWith` that used to pin the total arm's `where`. Only the embedding arm kept one.

**Fix — both arms, both mechanisms:**

- `usageRows` rows may now carry a `sessionId`, and the matcher honours it. An **absent** `sessionId` filter matches every row, exactly as the real query would, so a lost scope **over-counts here as it would in production** rather than quietly matching nothing.
- Both arms of the split now carry an explicit `toHaveBeenCalledWith` pinning every scoping key (`sessionId`, `estimatedCostUsd`, and `agentStep` on the embedding arm).
- A new case, `counts only THIS run's unpriced rows, never another run's (#94)`, seeds a sibling run's rows with the same phases and the same NULL cost. It matters because the LLM share is computed by **subtraction**: an unscoped total inflates `unpricedLlmTokens`, which is the field `exceeded()` stops a run on — so a sibling run's cold embedding pass could stop this one.

### Docs

`docs/USER_GUIDE.md` → *Test coverage → Budget guard* now states that the **indexing** pass is charged to the same `TESTCOVERAGE_BUDGET_CENTS` cap as the judge and suggestion phases, that a cold corpus on a *priced* cloud embedder can consume the cap before judging starts (`BUDGET_EXCEEDED` with almost nothing judged), and what to do about it. `docs/ARCHITECTURE.md` already documented the mechanism (#72); the operator-facing warning was the gap the shipped fragment left.

## Mutations — every test was watched go red

| # | Mutation | Went red |
|---|---|---|
| M1 | Restore the `projectIds.length > 0` hatch in `loadAccessibleImpactDetail` | 5 tests: detail read, markdown export, Jira publish, and both legacy-run cases |
| M2 | Detail `projectIds` back to item-derived only | `names the started-for projects of a run that has written no item yet`; `unions the started-for projects with the item-derived ones` |
| M3 | Drop `startedById` from the detail projection | `reports no projects, and the starting actor, for a run with neither` |
| M4 | Drop `actorId` from the list route call | all 4 `GET /` call-shape assertions |
| M5 | Drop the `startedById` fallback from the list filter (`return false`) | `shows a run with no recoverable projects to the actor who started it` |
| M6 | Make the fallback unconditional — reopen the hatch (`return true`) | `still hides it from every other member`; `hides it when no actor is supplied at all` |
| M7 | Default a missing starter to the caller: `(detail.startedById ?? actor.id) !== actor.id` | `is refused when the run names no starter` |
| M8 | Drop `assertProjectAccess` from `GET /sync/drift` | the 404 case and the argument-pinning case |
| M9 | Drop `assertProjectAccess` from `GET /sync/drift/count` | its 404 case |
| M10 | Assert access against a hardcoded `"proj-1"` instead of the requested id | `GET /sync/drift returns 404 for a project the caller cannot access` |
| M11/M12 | Pass an absent caller straight through the narrow on either drift read | `returns 401 rather than skipping the scope check when there is no caller` |
| M13 | **#94** — drop `sessionId` from the unpriced-**total** aggregate | `counts only THIS run's unpriced rows, never another run's (#94)` — `expected 9007100 to be 100` — and the new total-arm `where` pin |
| M14 | **#94** — drop `sessionId` from the unpriced-**embedding** aggregate | the same two — `expected 9000400 to be 400` |

Every mutation was reverted and the file confirmed byte-identical afterwards.

## Gates

| Gate | Result |
|---|---|
| `pnpm lint` | clean (whole repo, `--max-warnings=0`, plus the NUL / private-terms / licence / CLA verifiers) |
| `pnpm typecheck` | clean, all 8 packages |
| `pnpm test` | `1125 + 362 + 58 + 36 + 12 + 5 + 1` files passed, 0 failed |
| `pnpm changelog:verify` | passes — 2 fragments, no `CHANGELOG.md` edit |
| `pnpm agents:verify` / `pnpm skills:verify` | pass |
| `pnpm audit` | 4 low, 25 moderate, **0 high/critical** — all pre-existing; this PR adds no dependency |

### Coverage (`server`, v8, project floor 80% global on all four metrics)

| | before | after |
|---|---|---|
| **global** stmts / branch / func / lines | 92.16 / 83.72 / 92.87 / 93.45 | 92.16 / **83.73** / 92.87 / **93.46** |
| `routes/impact-analysis.ts` stmts / branch | 97.56 / 91.42 | **97.59 / 91.54** |
| `lib/impact-analysis/impact-analysis-read.ts` | 95.23 / 86.26 | **95.33 / 86.70** |
| `routes/sync.ts` | 83.16 / 66.66 | **85.04 / 68.65** |
| `lib/testcoverage/cost-tracker.ts` | 100 / 92.85 | 100 / 92.85 (unchanged — #94 is a test-side fix) |

Measured by running the full server suite with `--coverage` on `origin/main` and on this branch.

## Deliberately not done

- **`POST /sync/drift/:id/resolve` has the same class of gap.** It gates on `sync.resolve` and resolves the project from the drift row, never checking the actor against it. #88's acceptance criteria name only the two reads, so it is out of scope here — but it is a real A01 finding and wants its own issue.
- **`server/tsconfig.json` excludes `tests` and `src/**/*.test.ts`, so `pnpm typecheck` never type-checks server test files.** Adding a required field to `ImpactAnalysisDetail` left three annotated server fixtures silently type-incorrect; the UI's two were caught immediately because `ui` *does* typecheck its tests. I fixed all five, but the gate hole itself is pre-existing and not this change's job.
- No version bump; no `CHANGELOG.md` edit; `e2e/` and the `e2e` CI job untouched.


mgcronin (commented):
## Code Review — PR #101 (automated analysis, not a human sign-off)

Recommendation: merge with follow-ups — 0 blocking findings, 6 non-blocking.

The reviewing `gh` identity is also the PR author, so GitHub cannot record an independent approval or change request here. This is published as a comment; the merge decision stays with the caller.

### Requirements: PASS
**#88** — [x] detail/export/Jira-publish (and re-run and drift, five routes not three) authorize against the persisted `impact_analysis_projects` selection, with the empty-list hatch removed; [x] both `/sync/drift` reads call `assertProjectAccess` before the read, 404 not 403; [x] a legacy run with neither persisted projects nor items stays visible to its `startedById` actor on both the list and detail paths; [x] a test per item, each proven to fail before the fix.
**#94** — [x] dropping `sessionId` from the unpriced-total aggregate now turns two tests red; [x] both arms carry an explicit `toHaveBeenCalledWith` pinning every scoping key; [x] `docs/USER_GUIDE.md` documents that index-phase embedding spend counts against `TESTCOVERAGE_BUDGET_CENTS`.

### Falsifiability — I re-ran 5 of the 14 claimed mutations myself, each reverted afterwards (tree byte-identical)
| Mutation | Claimed | Observed |
|---|---|---|
| Reopen `projectIds.length === 0` → `return detail` | 5 red | **5 red** |
| Detail `projectIds` back to item-only | 2 red | **2 red** |
| List fallback → `return true` | 2 red | **2 red** |
| Drop `assertProjectAccess` from `GET /sync/drift` | 2 red | **2 red** |
| Drop `sessionId` from the unpriced-**total** aggregate | `expected 9007100 to be 100` + the new pin | **exactly that** |

The mutation table in the PR body is accurate as written — that is not always the case and is worth saying.

### Security (OWASP A01): the two reported gaps are genuinely closed
The order is right (scope check before the read, argument-pinned), the denial is 404 on every arm so there is no existence oracle, a projection that names no starter is refused rather than defaulted to the caller, and the inline `!req.user → 401` + `assertProjectAccess` shape mirrors `requireProjectAccess` exactly. Run creation writes `impact_analysis_projects` in the same `prisma.impactAnalysis.create` as the run (`impact-analysis-engine.ts:833`), so every run created since #70 is attributable — the starter fallback really is legacy-only. `GET /:id/publish/jira` on an unattributable run stops at `JIRA_NOT_CONFIGURED`, leaking nothing.

One genuine A01 remains adjacent and unfiled (`POST /sync/drift/:id/resolve`) — inline below.

### Tests: sufficient, with one falsifiability hole
104 tests pass across the five touched server files. The hole is that the `include` half of the #88 fix is pinned by nothing — measured, inline below.

### Performance / quality / docs
One extra relation on the detail query and one `project.findUnique` per drift read; no N+1, no unbounded read. `npx eslint` is clean on every file this PR touched. `pnpm changelog:verify` passes (two fragments, no `CHANGELOG.md` edit, no version bump). `docs/USER_GUIDE.md` gains the operator warning #94 asked for.

### CI
Plain re-read: 14 checks pass, 0 pending; `e2e` and `windows` `skipping`; `api` pass 23m57s with `api-outcome` pass. Note the green was measured on base `dbedcea`; `origin/main` has since advanced two commits (#84, #100). No file this PR touches is touched by either, and GitHub now reports `MERGEABLE` / `CLEAN`, so this is a note rather than a finding.

### Scope of this review
Read all 16 changed files plus their surrounding modules, re-ran the five touched server test suites, re-ran five mutations, linted the changed files, ran the changelog verifier, traced every caller of `getImpactAnalysisDetail`/`listImpactAnalyses` and every `:id` route in `impact-analysis.ts`. I did **not** independently re-measure the coverage table (the server threshold is global 80/80/80/80 and CI's gate passed) and did not exercise the running application.

server/src/lib/impact-analysis/impact-analysis-access.test.ts
```diff
@@ -0,0 +1,212 @@
...
+    affectedTables: [],
+    feedback: [],
+  };
+}
+
+/**
+ * A prisma double WITHOUT `codeSymbol`/`codeEdge`, so the #962 write-path pass
+ * is skipped and the assertions measure the projection alone.
+ */
+function detailPrisma(row: DetailRow | null) {
```
mgcronin, line 81: **Medium / high confidence — the half of the #88 fix that lives in the `include` is pinned by no test.**

I mutated `getImpactAnalysisDetail`'s query back to `include: ITEM_INCLUDE` (dropping `projects: { select: { projectId: true } }`) and ran every impact-area suite: **17 files, 456 tests, all green.** In production that line is the entire mechanism that populates `row.projects`, i.e. what makes an in-flight run attributable at all.

The reason is this double: it answers `findFirst` from `where.id` alone and returns `projects` regardless of whether the caller asked for them, so the test cannot observe the query shape. The behavioural assertions above therefore prove the *projection* arithmetic (verified — M2 goes red exactly as claimed) but not the query.

This fails **closed**, not open: without the include an in-flight run reports `projectIds: []` and becomes starter-only rather than public. So it is not a security hole, which is why this is not blocking. It is, though, precisely the "the assertion exists, nothing makes it fire" shape that #94 — the other half of this same bundle — was filed about, and the remedy you already applied on the cost-tracker side works here too:

```ts
expect(findFirst).toHaveBeenCalledWith({
  where: { id: "ia-88" },
  include: expect.objectContaining({ projects: { select: { projectId: true } } }),
});
```

Worth doing in this PR or a fast follow-up.

server/src/routes/sync.ts
```diff
@@ -194,6 +207,8 @@ export function syncDriftRouter(): Router {
         const projectId = req.query.projectId as string;
         if (!projectId)
           throw new AppError(400, "INVALID_REQUEST", "projectId query param required");
+        if (!req.user) throw new AppError(401, "AUTH_REQUIRED", "Authentication required");
+        await assertProjectAccess(req.user, projectId);
```
mgcronin, line 211: **High severity, pre-existing, correctly out of #88's acceptance criteria — but it should not leave the building unfiled.**

Both reads are now scoped. `POST /drift/:id/resolve`, ~15 lines below this one, is not: it gates on `sync.resolve`, resolves the project from the drift row, and calls `resolveDriftEvent(id, action, userId)` without ever checking the actor against that project. That is the same A01 shape you just closed, on the **write** side — any `sync.resolve` holder can adopt or reject another project's drift by id, and adopting mutates that project's requirements.

You flagged it in the PR body and #88's AC genuinely names only the two reads, so I am not asking you to widen this PR. I checked `gh issue list` and there is no issue for it — please file one before this merges, otherwise the only record is a PR description.

I confirmed the fix you did ship is falsifiable: dropping `assertProjectAccess` from `GET /drift` turns 2 tests red, and from `/drift/count` turns 1 red.

server/src/routes/sync.ts
```diff
@@ -165,11 +166,19 @@ export function syncDriftRouter(): Router {
...
+   * pass the canonical object-level scope seam (`assertProjectAccess`, the same
+   * predicate `requireProjectAccess` applies to `/projects/:projectId/*`).
+   * A non-member gets 404, never 403, so probing cannot enumerate projects.
    */
   r.get("/drift", requirePermission("sync.read"), async (req: Request, res: Response, next) => {
     try {
       const projectId = req.query.projectId as string;
       if (!projectId) throw new AppError(400, "INVALID_REQUEST", "projectId query param required");
+      if (!req.user) throw new AppError(401, "AUTH_REQUIRED", "Authentication required");
+      await assertProjectAccess(req.user, projectId);
```
mgcronin, line 181: nit (low): `req.query.projectId as string` is a cast, not a narrowing. With Express's default `qs` parser, `?projectId=a&projectId=b` yields an **array**, which passes the `if (!projectId)` guard and reaches `assertProjectAccess` → `prisma.project.findUnique({ where: { id: ["a","b"] } })` → a Prisma validation error surfacing as a 500 rather than a 400.

Not a bypass — it throws before `listDriftEvents` — so this is robustness, not security. The canonical middleware you are mirroring already handles it (`middleware/require-project-access.ts`: `Array.isArray(raw) ? raw[0] : raw`, then `String(...)`). Matching that here would make the two seams behave identically for a malformed request as well as a well-formed one. Same line in `/drift/count` below.

server/src/routes/impact-analysis.ts
```diff
@@ -428,11 +440,20 @@ async function loadAccessibleImpactDetail(
...
   }
-  if (!isAdminActor(actor) && detail.projectIds.length > 0) {
-    const accessible = new Set(await listAccessibleProjectIds(actor));
-    if (!detail.projectIds.every((pid) => accessible.has(pid))) {
+  if (isAdminActor(actor)) return detail;
+
+  if (detail.projectIds.length === 0) {
+    // Unattributable run: the starter only. A projection that names no starter
+    // matches no actor, so it is refused — never defaulted to the caller.
+    if (detail.startedById !== actor.id) {
```
mgcronin, line 448: Good: fail-closed by construction — a projection with no `startedById` matches no actor, and the admin arm returns before this, so there is no path where a missing starter defaults to the caller. I verified the M7-shaped mutation (`(detail.startedById ?? actor.id) !== actor.id`) is what the `is refused when the run names no starter` case exists to catch.

One consequence worth knowing rather than fixing (informational): the list path admits a run if **any** of its projects is accessible, while this path requires **all**. That asymmetry is pre-existing, but the fix now extends it to in-flight runs — a member of one project in a multi-project run will see the run listed and get a 404 opening it, where the old hatch gave them a 200. That is the correct security answer; it may be worth a UI-side message at some point so it does not read as a bug.

packages/shared/src/impact.ts
```diff
@@ -568,8 +568,20 @@ export interface ImpactAnalysisDetail {
...
+   * which is what lets the read routes authorize a run that has no items yet.
+   */
   projectIds: string[];
+  /**
+   * #88 — the actor who started the run. Load-bearing for authorization, not
+   * display: a pre-#70 run with neither persisted projects nor items belongs to
+   * no project the access filter can name, and its starter is then the only
+   * non-admin principal who may still read it.
+   */
+  startedById: string;
```
mgcronin, line 584: Informational: `startedById` is now on the wire for `GET /api/impact-analyses/:id`, exposing a user id to every member who can read the run, for a field whose only purpose is server-side authorization — the comment says as much ("load-bearing for authorization, not display").

Low risk (ids already appear in `impactAnalysisSchema` and the UI fixtures set it), and keeping one type for both layers is reasonable. Flagging only so the choice is deliberate: if nothing renders it, the route could strip it after `loadAccessibleImpactDetail` returns and the type could carry it as read-layer-only.

docs/USER_GUIDE.md
```diff
@@ -4440,6 +4440,18 @@ OpenAI-compatible endpoint before the judge had started. Price the model with a
 `MODEL_PRICES` key of the form `embed:<backend>:<model>` — for example
 `embed:bedrock-sdk:cohere.embed-english-v3` — to bring it back onto the cap.
 
+**Priced embedding spend from the indexing pass counts against the run's cap.**
```
mgcronin, line 4443: Good — this is exactly what #94's third criterion asked for, and it says what an operator should *do* (raise the cap for the first cold run, or use a local embedder) rather than only what happens.

Separate low doc nit, not on this file: `docs/ARCHITECTURE.md:2345-2347` states the impact-analysis rule as "detail reads are rejected unless **every** project in the run is accessible", which is now incomplete — it does not mention the starter-only exception for a run with no recoverable projects. One clause there would keep the architecture doc the source of truth for the authz rule you just changed.

mgcronin:
Follow-ups filed from the review before merge: #102 (the same A01 on the write side — `POST /sync/drift/:id/resolve` has no project scope) and #103 (the `include` half of #88's fix is pinned by no test; 456 tests stay green when it is reverted).

